PR #1, opened 2016-08-19
Add scheduling of morning briefings

tomrf1:
(and various improvements)

- The bot will push a morning briefing (a set of 5 links) to a user at a time specified by them (restricted to 6, 7 or 8am).
- The user's ID and preferences are stored in a dynamodb table.
- It takes the user's current timezone offset from facebook to calculate the correct time to push.
- At intervals of 15 mins, it checks for users in the dynamo table with the current time (all done in UTC). It checks at this point to see if the user's timezone has changed.

The user can request headlines (editors-picks) or trending (most-viewed) at any time. The front (uk/us/au/international) is determined using the facebook user's locale.

### DIFF
--- a/cloudformation/cloudformation.yml
+++ b/cloudformation/cloudformation.yml
@@ -59,13 +59,16 @@ Parameters:
         Description: Url of facebook messenger endpoint
         Type: String
         Default: https://graph.facebook.com/v2.6/me/messages
+    DynamodbTableName:
+        Description: Name of dynamodb table used to store user data
+        Type: String
 
 Mappings:
     StageMap:
         PROD:
-            MinSize: 3
-            MaxSize: 6
-            DesiredCapacity: 3
+            MinSize: 1
+            MaxSize: 2
+            DesiredCapacity: 1
             InstanceType: t2.micro
         CODE:
             MinSize: 1
@@ -116,6 +119,32 @@ Resources:
                         - ec2:DescribeInstances
                     Resource:
                         - "*"
+            Roles:
+                - Ref: FacebookNewsBotRole
+
+    DynamoPolicy:
+        Type: AWS::IAM::Policy
+        Properties:
+            PolicyName: dynamo-policy
+            PolicyDocument:
+                Statement:
+                    Effect: Allow
+                    Action:
+                        - dynamodb:GetItem
+                        - dynamodb:PutItem
+                        - dynamodb:UpdateItem
+                        - dynamodb:Query
+                    Resource:
+                        "Fn::Join":
+                            - ""
+                            -
+                                - "arn:aws:dynamodb:"
+                                - Ref: AWS::Region
+                                - ":"
+                                - Ref: AWS::AccountId
+                                - ":table/"
+                                - Ref: DynamodbTableName
+                                - "*"
             Roles:
                 - Ref: FacebookNewsBotRole
 
@@ -384,4 +413,6 @@ Resources:
                                     - Ref: FacebookUrl
                                     - " DEFAULT_IMAGE_URL="
                                     - Ref: DefaultImageUrl
+                                    - " DYNAMODB_TABLE_NAME="
+                                    - Ref: DynamodbTableName
                                     - " pm2 start application.pm2.json"
--- a/cloudformation/dynamodb.yml
+++ b/cloudformation/dynamodb.yml
@@ -1,0 +1,37 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: DynamoDB Tables Template
+Parameters:
+  Stage:
+    Description: Stage name
+    Type: String
+    AllowedValues:
+      - CODE
+      - PROD
+    Default: CODE
+Resources:
+  facebookNewsBotUsers:
+    Type: AWS::DynamoDB::Table
+    Properties:
+      AttributeDefinitions:
+        - AttributeName: ID
+          AttributeType: S
+        - AttributeName: notificationTimeUTC
+          AttributeType: S
+      KeySchema:
+        - AttributeName: ID
+          KeyType: HASH
+      GlobalSecondaryIndexes:
+        - IndexName: notificationTimeUTC-ID-index
+          KeySchema:
+            - AttributeName: notificationTimeUTC
+              KeyType: HASH
+            - AttributeName: ID
+              KeyType: RANGE
+          Projection:
+            ProjectionType: ALL
+          ProvisionedThroughput:
+            ReadCapacityUnits: '5'
+            WriteCapacityUnits: '1'
+      ProvisionedThroughput:
+        ReadCapacityUnits: '5'
+        WriteCapacityUnits: '1'
--- a/index.js
+++ b/index.js
@@ -1,19 +1,18 @@
 "use strict"
 const Express = require("express")
 const BodyParser = require("body-parser")
-const Request = require('request')
+const Request = require("request")
+const Moment = require("moment")
 const UserStore = new (require("./lib/user-store"))
-const get = require('simple-get-promise').get
-const asJson = require('simple-get-promise').asJson
+const Capi = require("./lib/capi")
+const Facebook = require("./lib/facebook")
+const Schedule = require("node-schedule")
 
 const VERIFY_TOKEN = process.env.VERIFY_TOKEN
 const ACCESS_TOKEN = process.env.ACCESS_TOKEN
-const API_KEY = process.env.API_KEY
 
 const DEFAULT_IMAGE_URL = process.env.DEFAULT_IMAGE_URL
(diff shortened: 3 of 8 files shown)


cloudformation/dynamodb.yml
```diff
@@ -0,0 +1,44 @@
...
+              KeyType: RANGE
+          Projection:
+            ProjectionType: ALL
+          ProvisionedThroughput:
+            ReadCapacityUnits: '10'
+            WriteCapacityUnits: '10'
+      ProvisionedThroughput:
+        ReadCapacityUnits: '10'
+        WriteCapacityUnits: '10'
+      TableName:
```
cb372: Something to be aware of: specifying a table name means you are heavily restricted in what kind of updates you can do using CloudFormation. e.g. if you want to add a new index or something like that, you will have to do it manually, so your table will then be out of sync with the CloudFormation file.

cloudformation/cloudformation.yml
```diff
@@ -59,13 +59,16 @@ Parameters:
...
 Mappings:
     StageMap:
         PROD:
-            MinSize: 3
-            MaxSize: 6
-            DesiredCapacity: 3
+            MinSize: 1
+            MaxSize: 2
+            DesiredCapacity: 1
             InstanceType: t2.micro
```
LATaylor-guardian: we know we have difficulty running jvm based apps on a `nano` but this might be fine here.

index.js
```diff
@@ -23,112 +22,194 @@ app.use(BodyParser.json())
...
+    })
   },
-  subscribe(id, time) {
-    UserStore.setNotificationTime(id, time)
-    sendMessage(
+  morningBriefing(id) {
+    Facebook.sendTextMessage(
       id,
-      buildMenu("Fantastic. You can change your subscription at anytime by asking for 'help' or 'menu'.\nWould you like the headlines or trending news?")
+      "Good morning from the Guardian. Here are today's headlines:",
```
tsop14: Should this say, "here is the morning briefing"? `"I can give you the headlines, trending news or deliver a morning briefing to you."` suggests they are different but here they appear the same thing, which would confuse me.

cloudformation/cloudformation.yml
```diff
@@ -384,4 +413,6 @@ Resources:
                                     - Ref: FacebookUrl
                                     - " DEFAULT_IMAGE_URL="
                                     - Ref: DefaultImageUrl
+                                    - " DYNAMODB_TABLE_NAME="
+                                    - Ref: DynamodbTableName
```
LATaylor-guardian: What does the table name look like? Can it be constructed from already passed in parameters such as stack and stage, making the rest up with hard coded strings?

cloudformation/dynamodb.yml
```diff
@@ -0,0 +1,44 @@
...
+        - IndexName: notificationTimeUTC-ID-index
+          KeySchema:
+            - AttributeName: notificationTimeUTC
+              KeyType: HASH
+            - AttributeName: ID
+              KeyType: RANGE
+          Projection:
+            ProjectionType: ALL
+          ProvisionedThroughput:
+            ReadCapacityUnits: '10'
```
LATaylor-guardian: Its worth starting off with having these lower to reduce cost and scaling up as needed (unless you've investigated this already, then ignore me)

index.js
```diff
@@ -266,34 +379,39 @@ app.post("/webhook/", (req, res) => {
...
-          Events.start(id)
-        }
+        Events.start(id)
       } else if (text.match(/^help *|^what can you do */)) {
-        Events["help"](id)
+        Events.help(id)
       } else if (text.match(/^menu */)) {
-        Events["menu"](id)
+        Events.menu(id)
+      } else if (text.match(/^subscribe */)) {
```
cb372: How do all these regexes with a space at the end match against the text (which you have trimmed)?

index.js
```diff
@@ -23,112 +22,194 @@ app.use(BodyParser.json())
...
-  )
+function greetNewUser(id, name) {
+  UserStore.addUser(id, (err, data) => {
+    if (err) {
+      console.log("Error adding user "+ id +": "+ JSON.stringify(err))
+    } else {
+      Facebook.sendMessage(
+        id,
+        buildButtonsAttachment(
+          "Hi there "+ name+ ", I'm a virtual assistant created by the Guardian to keep you up-to-date with the latest news.\n\nCan I deliver you the headlines each morning?",
```
tsop14: "Can I deliver you the headlines each morning?" Also think this is confusing. The morning briefing isn't necessarily the same as the headlines at the moment?

index.js
```diff
@@ -23,112 +22,194 @@ app.use(BodyParser.json())
 
 const Events = {
   unknown(id) {
-    sendMessage(
+    Facebook.sendMessage(
       id,
       buildButtonsAttachment(
-        "Sorry! I’m pretty stupid at the moment. My creators are training me to understand and do more.",
+        "Sorry! I'm pretty stupid at the moment. My creators are training me to understand and do more.",
```
LATaylor-guardian: I'm not sure about the implication of AI  that 'training' gives here.

index.js
```diff
@@ -23,112 +22,194 @@ app.use(BodyParser.json())
...
-      buildMenu("Fantastic. You can change your subscription at anytime by asking for 'help' or 'menu'.\nWould you like the headlines or trending news?")
+      "Good morning from the Guardian. Here are today's headlines:",
+      Events.headlines(id)
     )
   },
   headlines(id) {
-    getEditorsPicks(id)
+    Facebook.getFacebookUser(id, (error, response, body) => {
+      const front = localeToFront(JSON.parse(body)["locale"])
+      Capi.getEditorsPicks(id, front, sendCapiResults)
```
LATaylor-guardian: What happens if the request fails in this case?

index.js
```diff
@@ -238,11 +301,61 @@ function getImageUrl(item) {
...
   })
+  if (typeof mainElement !== "undefined") {
+    return mainElement
+  } else {
+    return elements[0]
+  }
 }
 
+/*
+ *  Every 15 mins (the smallest interval between timezones), scan the dynamo table
```
cb372: (actually a query, not a scan)

tsop14:
Added some comments on the language used. Nice to have separated facebook/capi code - much better bot!

index.js
```diff
@@ -182,48 +263,30 @@ function buildGenericAttachment(elements) {
...
 // Look for image with width 1000, otherwise the next widest
 function getImageUrl(item) {
   let imageUrl = DEFAULT_IMAGE_URL
   let widest = 0
   if ("elements" in item) {
-    let element = getMainElement(item.elements)
+    const element = getMainElement(item.elements)
     if (typeof element !== "undefined") {
       if ("assets" in element && element.assets.length > 0) {
         for (let assetIdx in element.assets) {
```
LATaylor-guardian: element.assets.forEach() ?

index.js
```diff
@@ -238,11 +301,61 @@ function getImageUrl(item) {
...
 }
 function getMainElement(elements) {
-  return elements.find(element => {
+  const mainElement = elements.find(element => {
     return ("relation" in element && element.relation === "main")
   })
+  if (typeof mainElement !== "undefined") {
+    return mainElement
+  } else {
+    return elements[0]
```
LATaylor-guardian: Is it possible for item at elements[0] to not exist here?

index.js
tomrf1: I thought that as well. I think i'll get rid of it...

index.js
```diff
@@ -238,11 +301,61 @@ function getImageUrl(item) {
...
+  }
 }
 
+/*
+ *  Every 15 mins (the smallest interval between timezones), scan the dynamo table
+ *  for users with a notificationTimeUTC equal to current UTC time, and push to them.
+ *  Also check if their timezone has changed since the user was last handled, and
+ *  update if so.
+ */
+const recRule = new Schedule.RecurrenceRule()
```
LATaylor-guardian: this code looks like it could live outside of this file.

index.js
```diff
@@ -238,11 +301,61 @@ function getImageUrl(item) {
   return imageUrl
 }
 function getMainElement(elements) {
-  return elements.find(element => {
+  const mainElement = elements.find(element => {
     return ("relation" in element && element.relation === "main")
   })
+  if (typeof mainElement !== "undefined") {
```
LATaylor-guardian: oh Javascript....

index.js
tomrf1: yes, if there are no body elements either, but the calling function does check for this.

cloudformation/cloudformation.yml
tomrf1: after seeing @cb372 's comment about table names I think I will let cloudformation generate the table name

index.js
chriswilk: I'm going to do another iteration of the text / speech very soon.

PR #2, opened 2016-08-22
Add 'more' button after links

tomrf1:
A "quick reply" button to page through the results (headlines or trending).
Also changed the payload string to be json.


LATaylor-guardian:
What does a "quick reply" button look like and how does this change from what we currently have?

tomrf1:
Currently there is no ability to get the next 5 results.
Quick reply is documented here https://developers.facebook.com/docs/messenger-platform/send-api-reference/quick-replies
It's essentially a differently-styled button that appears over the text input

LATaylor-guardian:
We like these buttons 👍

PR #4, opened 2016-08-23
Randomise messages

tomrf1:
@chriswilk


chriswilk:
Looks good to me 👍

LATaylor-guardian:
@chriswilk Why are we randomising messages for?

PR #7, opened 2016-08-23
Fix processing of negative timezone offsets

tomrf1:
It ignored our first American user :(

### DIFF
--- a/index.js
+++ b/index.js
@@ -198,7 +198,7 @@ function localeToFront(locale) {
 //Returns the time adjusted by the offset as a string
 function getUTCTime(time, offset) {
   //facebook timezone offset is a float in [-24, 24]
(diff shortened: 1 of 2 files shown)


LATaylor-guardian:
👍

index.js
```diff
@@ -198,7 +198,7 @@ function localeToFront(locale) {
 //Returns the time adjusted by the offset as a string
 function getUTCTime(time, offset) {
   //facebook timezone offset is a float in [-24, 24]
```
cb372: Not [-12, 12]? If your timezone was -13, you'd be so far in past you end up in the future!

tomrf1: :/ https://developers.facebook.com/docs/graph-api/reference/user

PR #8, opened 2016-08-24
Make edition configurable

tomrf1:
We decide which front to use by checking the user's facebook locale string (e.g. "en_GB" => "uk")
It appears this never changes, so e.g. a British person who moved to the US still has "en_GB" (which makes sense).
We'll still use this as a default for now, but the user can now change it by clicking "Change edition". The options are presented as a carousel (like with the headlines), because you're only allowed a max of 3 buttons. In future we could add images for each tile in the carousel.

Note - we have plans to add bots to the US and AU facebook pages as well. The default edition can then be specific to the facebook page.

picture 2

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,7 +47,7 @@ const Events = {
   },
   start(id) {
     //Is this an existing user?
-    UserStore.getUser(id, (err, data) => {
+    UserStore.getUser(id, (err, dynamoData) => {
       if (err) {
         console.log("Error looking up user "+ id +": "+ JSON.stringify(err))
       } else {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -446,17 +485,15 @@ app.post("/webhook/", (req, res) => {
     if (event.message && event.message.text) {
       const text = event.message.text.trim().toLowerCase()
 
-      if (text === "more stories") {
+      if (typeof event.message.quick_reply !== "undefined") {
```
tsop14: Could there be a type error if e.g. event.message is `undefined`?

tomrf1: At this point we know `event.message` is defined

tsop14:
I don't really understand the difference between the editions. If you get chose AU front would you only get Australian news? Or just prioritised Australian news? I think an explanation might be nice so it is clear what "change edition" will do.

PR #10, opened 2016-08-26
Always lookup user in dynamo

tomrf1:
New users should first be presented with a 'Get started' button (this is a standard facebook messenger thing - https://developers.facebook.com/docs/messenger-platform/thread-settings/get-started-button).
This triggers the `start` event, which adds the user to dynamodb and greets them.

However, the Get started button keeps disappearing from facebook messenger (@chriswilk has asked facebook to sort this out). We should defend against this by always looking up the user in dynamo whenever a message is received from facebook. So even if they haven't arrived via the Get started button, we'll add them to dynamo and send the initial greeting.

### DIFF
--- a/index.js
+++ b/index.js
@@ -526,7 +526,16 @@ function handleExistingUser(id, event, userData) {
     log(id, "postback", payload.event, {"payload": payload})
 
     if (Events[payload.event]) {
-      Events[payload.event](userData, payload)
+      if (payload.event === "start") {
+        /*
+         * This shouldn't generally happen, but it's possible for an existing
+         * user to get the 'Get started' button in a web browser if they
+         * delete the conversation first.
+         */
+        Events.greeting(userData)
+      } else {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -342,7 +329,7 @@ function buildGenericAttachment(elements) {
   }
 }
 
-function getAndSendCapiResults(id, type, page) {
+function getAndSendCapiResults(user, type, page) {
   const sendCapiResults = (id, results) => {
```
tsop14: Does `id` now need to be `user`?

tsop14: Should `id` now be `user`?

tomrf1: Here `id` is the name of a parameter to the function assigned to `sendCapiResults`. So `id` is passed in by the `capi` library when it calls that function.
However, it is kind of pointless having the `id` parameter, since it's available here as `user.ID`, so I'll remove it now.

tsop14:
👍

PR #11, opened 2016-08-26
Fix initial menu for new users

tomrf1:
Immediately after a new user subscribes for the first time, a menu button incorrectly says "Subscribe" rather than "Manage subscription". This fixes that.

(also added logging for the `start` event)


chriswilk:
👍

PR #13, opened 2016-08-30
Use item_url on elements and add 'share' button

tomrf1:
### item_url
Setting this turns each tile on the carousel into a link, which will open in Instant Articles if available there. This means we can remove the "Read story" button.

### Share button
This is a bit of a hack, since FB doesn't provide a way of sharing links from the carousel. A "Share" button appears below each tile, and when clicked we send a link to the article as a text message. Messenger automatically accompanies that link with a share button, allowing the user to share on Messenger only.
The URL has the campaign code `fb_newsbot_share`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -342,7 +342,7 @@ function getAndSendCapiResults(user, type, page) {
       return buildElement(
         item.webTitle, [
           buildButton("postback", "Share", buildPayload("share", {
-            "url": item.webUrl,
+            "url": item.webUrl + "?"+ CAMPAIGN_CODE_PARAM +"_share",
             "title": item.webTitle
           }))
         ],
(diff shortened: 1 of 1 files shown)


LATaylor-guardian:
By 'text message' do you mean SMS? Or a message back to the user in the messenger window?

LATaylor-guardian:
I'm guessing the latter now I read it more - ignore me.

tomrf1:
Yeah I meant just a basic message containing text - no attachments, buttons etc. "Text message" is a bit of a loaded term isn't it

LATaylor-guardian:
Annoying they don't provide a nicer way to share - hopefully we can feed this back to them and they can improve @chriswilk.

👍

PR #15, opened 2016-08-30
Cache capi results

tomrf1:
Currently the app sends a CAPI request every time it needs the editors-picks or most-viewed. This is not at all scalable, especially in the case of the morning briefing.

Cached results are considered fresh for 30 seconds.

### DIFF
--- a/lib/capi.js
+++ b/lib/capi.js
@@ -19,9 +19,9 @@ function Cache() {
   this._data = {}
 }
 Cache.prototype.get = function(url) {
-  const data = this._data[url]
-  if (typeof data !== "undefined" && Moment.utc().diff(data.timestamp) < data.staleMs) {
-      return this._data[url]
+  const item = this._data[url]
+  if (typeof item !== "undefined" && Moment.utc().diff(item.timestamp) < item.staleMs) {
+      return item.data
   } else {
       return null
   }
@@ -51,7 +51,7 @@ function doCapiQuery(id, type, front, callback) {
 
   const cached = CACHE.get(url)
   if (cached !== null) {
-    callback(cached.data)
+    callback(cached)
   } else {
     get(url)
       .then(asJson)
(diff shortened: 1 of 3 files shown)


lib/capi.js
```diff
@@ -7,10 +7,27 @@ module.exports = {
...
+}
+Cache.prototype.get = function(url) {
+  return this._data[url]
+}
+Cache.prototype.put = function(url, data) {
+  this._data[url] = {
+    "timestamp": Moment.utc(),
+    "data": data
+  }
+}
```
cb372: There is [node-cache](https://www.npmjs.com/package/node-cache) but maybe it's overkill?

lib/capi.js
```diff
@@ -25,13 +42,24 @@ function doCapiQuery(id, type, front, callback) {
...
-  get(url)
-    .then(asJson)
-    .then(json => {
-      const fieldName = type === "editors-picks" ? "editorsPicks" : "mostViewed"
-      callback(json.response[fieldName])
-    })
-    .catch(error => { console.log("CAPI error: "+ error) })
+
+  const cached = CACHE.get(url)
+  if (typeof cached !== "undefined" && Moment.utc().diff(cached.timestamp) < CACHE_STALE_MILLISECONDS) {
```
cb372: You could do the TTL check inside `Cache.get` instead of leaking it to here. Also `Cache.put` could take a TTL as an optional 3rd argument, so you can cache different things with different TTLs.

lib/capi.js
```diff
@@ -19,12 +19,18 @@ function Cache() {
   this._data = {}
 }
 Cache.prototype.get = function(url) {
-  return this._data[url]
+  const data = this._data[url]
+  if (typeof data !== "undefined" && Moment.utc().diff(data.timestamp) < data.staleMs) {
+      return this._data[url]
```
cb372: Why not return `data.data`? The client only cares about that, not internals like `timestamp` and `staleMs`.

cb372:
👍

PR #17, opened 2016-08-31
Fix handling of missing timezone in facebook response

tomrf1:
This caused people to miss their morning briefings this morning :(
Before sending a morning briefing, it checks if the user's timezone has changed (according to facebook). It turns out sometimes this field is missing in the facebook response - if this happens it now just pushes the morning briefing anyway.

This is a good reason for switching to scalajs - the bug is the result of an incorrect variable name in an error-handling branch after refactoring!

### DIFF
--- a/index.js
+++ b/index.js
@@ -464,7 +464,9 @@ function notifyUser(dynamoData, fbData) {
       })
     }
   } else {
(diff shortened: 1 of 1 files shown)


cb372:
👍

PR #18, opened 2016-09-01
Include campaign code

tomrf1:
It lost the campaign code after switching from using buttons to item_urls
@chriswilk


chriswilk:
+1

PR #20, opened 2016-09-13
Refactor

tomrf1:
index.js was getting quite big, so I've moved most of it out into events.js, scheduler.js, and helpers.js.
This is in preparation for adding some new functionality.

I've tested everything in CODE


tsop14:
👍 tidy!

PR #21, opened 2016-09-14
Stories by topic

tomrf1:
1. By headline, e.g. "football headlines"
2. By most popular, e.g. "popular football stories"
3. Include "quick replies" after the carousel, to suggest some topics

For 'topics', I've used the same definitions that we used for alexa.
For some topics we use `editors-picks`, for others the `results` field. Either way, it always gets up to 25 items and caches them.

picture 3


lib/capi.js
```diff
@@ -34,38 +37,516 @@ Cache.prototype.put = function(url, data, staleMs) {
...
+function isTopic(text) {
+  return queryProperties(text) ? true : false
 }
 
-function getMostViewed(id, front, callback) {
-  doCapiQuery(id, "most-viewed", front, callback)
+function getHeadlines(id, front, offset, topic, callback) {
+  const props = getQueryProperties(front, topic)
+
+  if (props !== null) {
```
LATaylor-guardian: Can this be simply  `if (props)`? null is not truthy.

lib/capi.js
```diff
@@ -34,38 +37,516 @@ Cache.prototype.put = function(url, data, staleMs) {
...
+    const url = CAPI_BASE_URL + props.path + "?"
```
LATaylor-guardian: The query above and below is pretty much the same. Can we build it in such a way that it's more reusable? (This might notg be of the uttermost importance now but will be if we continue to define new ways of querying). I think someone even contributed a JS CAPI client... maybe you can use that?

lib/capi.js
```diff
@@ -34,38 +37,516 @@ Cache.prototype.put = function(url, data, staleMs) {
...
 
-function doCapiQuery(id, type, front, callback) {
-  const url = CAPI_BASE_URL + 
-              front + 
-              "?page-size=0&show-"+ type +"=true"+
-              "&show-elements=image&show-fields=standfirst,thumbnail"+
-              "&api-key="+ API_KEY
+function getMostViewed(id, front, offset, topic, callback) {
+  const props = getQueryProperties(front, topic)
+  if (props !== null) {
```
LATaylor-guardian: As above.

lib/capi.js
```diff
@@ -34,38 +37,516 @@ Cache.prototype.put = function(url, data, staleMs) {
...
+      + (props.mostViewed ? "&show-most-viewed=true" : "&page-size="+ MAX_LINKS)
+    doCapiQuery(id, url, offset, callback)
+  } else {
+    console.log(`Error getting most viewed - cannot find properties object for ${front} / ${topic} for user ${id}`)
+  }
+}
 
+function doCapiQuery(id, url, offset, callback) {
   const cached = CACHE.get(url)
   if (cached !== null) {
```
LATaylor-guardian: And again.

lib/capi.js
```diff
@@ -34,38 +37,516 @@ Cache.prototype.put = function(url, data, staleMs) {
...
-    callback(cached)
+    callback(cached.slice(offset,offset+LINK_COUNT))
```
LATaylor-guardian: It's not clear to me what this is doing ` cached.slice(offset,offset+LINK_COUNT) `... maybe pull it out into a named variable.

lib/capi.js
```diff
@@ -34,38 +37,516 @@ Cache.prototype.put = function(url, data, staleMs) {
...
-  return type === "editors-picks" ? "editorsPicks" : "mostViewed"
+const getResults = (response) => {
+  if (response["editorsPicks"]) return response["editorsPicks"]
+  else if (response["mostViewed"]) return response["mostViewed"]
+  else return response["results"]
+}
+
+const getQueryProperties = (front, topic) => {
+  const props = queryProperties(topic ? topic : front)
+  if (props !== null) {
```
LATaylor-guardian: As above.

lib/capi.js
```diff
@@ -34,38 +37,516 @@ Cache.prototype.put = function(url, data, staleMs) {
...
+    else return props.default
+  } else return null
 }
 
+/**
+ * Get the properties object for a topic. This defines how to build the capi query.
+ * Each properties object can contain multiple objects for different editions, but it must have a 'default'
+ * object.
+ */
+const queryProperties = (topic) => {
```
LATaylor-guardian: Do you think this belongs in this file? Maybe better pulling it out?

lib/capi.js
tomrf1: I'll just do a simple refactor for now, but will look at using that JS client later - good point.

lib/capi.js
tomrf1: I did wonder about this, because it's enormous. But it's so specific to the internals of CAPI that it felt wrong putting it anywhere other than capi.js!

LATaylor-guardian: Yeah good point.

LATaylor-guardian:
👍

PR #23, opened 2016-09-20
Limit facebook messenger requests to 100/second

tomrf1:
FB have been rejecting some of our messenger api requests during the morning briefings. The error response from FB is vague, but our contact at FB suggested limiting it to 100 per second


tsop14:
👍

mchv:
@tomrf1 what does that mean for the end user in both case? If the message is rejected then it does not receive a response? if we rate limit ourself, does it means it add latency for users?

tomrf1:
Yes, if FB reject the message we don't retry, so the user receives nothing.
And yes it may add latency, but I wouldn't expect it to be noticeable to the user.

PR #25, opened 2016-09-29
Begin rewriting in scala with akka http

tomrf1:
I've created a separate branch, `v2`, from master.
From now on I will be merging the scala changes to `v2`. Once it's finished, I'll merge `v2` into master.

This is an initial outline of the app - it's very unfinished!

Notes:
- Messages from facebook may contain a text message or a payload.
- We store user notification time + their state in dynamodb
- On receiving a message from a user, we decide their next state and which messages to send to them based on their current state and the received message.


project/plugins.sbt
```diff
@@ -0,0 +1,3 @@
+addSbtPlugin("io.spray" % "sbt-revolver" % "0.8.0")
+
+addSbtPlugin("com.localytics" % "sbt-dynamodb" % "1.4.1")
```
LATaylor-guardian: woo!

src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -0,0 +1,91 @@
...
+import com.typesafe.config.Config
+import de.heikoseeberger.akkahttpcirce.CirceSupport
+import io.circe.generic.auto._
+import util.AWSConstants._
+import com.gu.cm.{Configuration, Mode}
+import com.gu.facebook_news_bot.logic.StateHandler
+
+import scala.concurrent.ExecutionContextExecutor
+
+trait BotService extends CirceSupport {
```
LATaylor-guardian: CirceSupport is cool.

src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -0,0 +1,91 @@
...
+  implicit val system: ActorSystem
+  implicit def executor: ExecutionContextExecutor
+  implicit val materializer: Materializer
+  val logger: LoggingAdapter
+
+  val config: Config = Configuration("facebook-news-bot.properties", Mode.Dev).load
```
LATaylor-guardian: All of this config can be pulled out into a Config object.

src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  val stage = config.getString("stage")
+
+  val userStore = {
+    val usersTable = config.getString("aws.dynamo.usersTableName")
```
LATaylor-guardian: What does getString return, a String or an Option[String]? If it returns a String you may want to use a function that returns an Option[String] and fail fast (sys.error) on the cases where important config to the app is missing.

src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -0,0 +1,91 @@
...
+    val client: AmazonDynamoDBClient = {
+      if (stage == "DEV") {
+        new AmazonDynamoDBClient().withEndpoint("http://localhost:8000")
+      } else {
+        val awsRegion = Regions.fromName(config.getString("aws.region"))
```
LATaylor-guardian: This can default to eu-west-1 IF it can't retrieve the region from config.

src/main/scala/com/gu/facebook_news_bot/util/AWSConstants.scala
```diff
@@ -0,0 +1,15 @@
+package com.gu.facebook_news_bot.util
+
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
+import com.amazonaws.auth._
+
+object AWSConstants {
+
+  val CredentialsProvider = new AWSCredentialsProviderChain(
```
LATaylor-guardian: new EnvironmentVariableCredentialsProvider() and new SystemPropertiesCredentialsProvider() arn't needed.

LATaylor-guardian: Also, this can probably be moved into your newly created Config object.

src/main/scala/com/gu/facebook_news_bot/BotService.scala
tomrf1: I left it as is because it does fail on startup with a `com.typesafe.config.ConfigException`, telling you which key is missing. Not sure handling it and doing a sys.error is necessary?

src/main/scala/com/gu/facebook_news_bot/services/UserStore.scala
```diff
@@ -0,0 +1,21 @@
+package com.gu.facebook_news_bot.services
+
+import com.amazonaws.services.dynamodbv2.document._
+import com.gu.facebook_news_bot.models.User
+
+class UserStore(db: DynamoDB, usersTableName: String) {
```
LATaylor-guardian: You should use https://github.com/guardian/scanamo for interacting with DynamoDB. It's written by Phil Wills and it's really great! Plus for your small use case it's a nice intro to it.

LATaylor-guardian: Maybe this class can move to a stores package rather than services?

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
```diff
@@ -0,0 +1,62 @@
...
+                        title: Option[String],
+                        payload: Option[String],
+                        image_url: Option[String])
+
+}
+
+/**
+  * Messages received from Messenger users
+  */
+case class FromFacebook(entry: List[FromFacebook.Entry])
```
LATaylor-guardian: FacebookUserMessage maybe?

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
```diff
@@ -0,0 +1,62 @@
...
+                     text: Option[String],
+                     attachments: Option[List[Attachment]],
+                     quick_reply: Option[QuickReply])
+
+  case class Attachment(`type`: String, payload: Payload)
+
+  case class Payload(url: Option[String])
+
+  //quick_reply and postback payloads contain the string that we supplied - this is different from the attachment payloads.
+  case class QuickReply(payload: String)
```
LATaylor-guardian: I think this and `Postback` below are a good example of when a value class is useful: 
http://docs.scala-lang.org/overviews/core/value-classes.html

src/main/scala/com/gu/facebook_news_bot/services/UserStore.scala
tomrf1: good idea, it does look nice!

mchv:
@tomrf1 if using `DynamoDB` to store the state I think you should use `Strong Consistent Read`

> When you request a strongly consistent read, DynamoDB returns a response with the most up-to-date data, reflecting the updates from all prior write operations that were successful. Note that a strongly consistent read might not be available in the case of a network delay or outage.

Otherwise it means the state you receive may not the be the last one, which has an end user is quite perturbing

src/main/scala/com/gu/facebook_news_bot/BotService.scala
LATaylor-guardian: Depends on the underlining library I guess. As long as it blows up correctly and stops the app from working.

src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  val routes = path("status") {
+    get {
+      complete("OK")
+    }
+  } ~ path("webhook") {
+    post {
+      entity(as[FromFacebook]) { fromFb =>
+        complete {
+          /**
```
LATaylor-guardian: As this code invoked by the route grows it might be nice to pull it out into a service layer that the routes call into.

src/main/scala/com/gu/facebook_news_bot/logic/StateHandler.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+    * @param userOpt  user from dynamodb
+    * @param message  message from Facebook messenger
+    * @return updated user, plus any messages to send to Facebook
+    */
+  def process(userOpt: Option[User], message: FromFacebook.Messaging): Future[Result] = {
+    val user = userOpt.getOrElse(newUser(message.sender.id))
+
+    val state = getStateFromString(user.state)
```
LATaylor-guardian: I find this a little confusing - what is `state`?

LATaylor-guardian: Ah I get it, so it's whether a user is new or returning?

tomrf1:
@mchv thanks. I think we may also need to use [this](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.OptimisticLocking.html) if it's to scale horizontally

src/main/scala/com/gu/facebook_news_bot/logic/StateHandler.scala
tomrf1: we store a `state` string in dynamo for each user. There could be many different states later on - at the moment I've just defined one for new users and one 'main' one.
It's necessary to have state in dynamo because FB doesn't allow us to attach persistent state on their side.

LATaylor-guardian:
> @mchv thanks. I think we may also need to use this if it's to scale horizontally

Yes, this problem could possibly exist at current.

LATaylor-guardian:
Looks like a good start!

tomrf1:
@LATaylor-guardian I've pushed a few commits with suggested changes.
configuration-magic expects you to already know your stage, so I've gone with an environment variable.

src/main/scala/com/gu/facebook_news_bot/BotConfig.scala
```diff
@@ -0,0 +1,49 @@
...
+  object Facebook {
+    val url = getString("facebook.url")
+    val accessToken = getString("facebook.accessToken")
+  }
+
+  object Capi {
+    val key = getString("capi.key")
+  }
+
+  private def getString(name: String): String = {
```
LATaylor-guardian: `getMandatoryString` maybe?

src/main/scala/com/gu/facebook_news_bot/BotConfig.scala
```diff
@@ -0,0 +1,49 @@
...
+    Try(config.getString(name)).getOrElse(sys.error(s"Error - missing mandatory config item, $name"))
+  }
+  private def getStringOrDefault(name: String, default: String): String = {
+    Try(config.getString(name)).getOrElse(default)
+  }
+  private def getInt(name: String): Int = {
```
LATaylor-guardian: As above.

src/main/scala/com/gu/facebook_news_bot/BotConfig.scala
```diff
@@ -0,0 +1,49 @@
...
+  val stage: Mode = sys.env.getOrElse("STAGE", "DEV") match {
+    case "DEV" => Mode.Dev
+    case _ => Mode.Prod   //CODE or PROD
+  }
+
+  val config: Config = Configuration("facebook-news-bot.properties", stage).load
+
+  val port = config.getInt("port")
+
+  object Aws {
```
LATaylor-guardian: As these are objects inside another object it's convention for them to be lowercased.

LATaylor-guardian:
Apart from comments 👍

PR #26, opened 2016-10-03
Implement UserStore

tomrf1:
Update and retrieve users from dynamodb.

I've added a `version` field to prevent data being overwritten when 2 messages from the same user are being processed concurrently. This effectively does the same thing as [optimistic locking](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.OptimisticLocking.html), but that uses java annotations and probably won't work with scanamo, plus it's extremely easy to implement here using conditional updates.

I still need to add [read consistency](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html), but I first need to add support to scanamo.


src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -26,19 +27,19 @@ trait BotService extends CirceSupport {
...
-        new AmazonDynamoDBClient().withEndpoint("http://localhost:8000")
+        new AmazonDynamoDBAsyncClient().withEndpoint("http://localhost:8000")
       } else {
         val awsRegion = Regions.fromName(BotConfig.aws.region)
-        new AmazonDynamoDBClient(BotConfig.aws.CredentialsProvider).withRegion(awsRegion)
+        new AmazonDynamoDBAsyncClient(BotConfig.aws.CredentialsProvider).withRegion(awsRegion)
       }
     }
-    new UserStore(new DynamoDB(client), usersTable)
+    new UserStore(client, usersTable, logger)
```
LATaylor-guardian: Can you use the Typesafe logger which provides the `StrictLogging` trait that can be extended so you don't have to pass logger around?

src/main/scala/com/gu/facebook_news_bot/models/User.scala
```diff
@@ -3,9 +3,10 @@ package com.gu.facebook_news_bot.models
 /**
   * The user data as it is stored in dynamodb
   */
-case class User(id: String,
+case class User(ID: String,
```
LATaylor-guardian: Why is this field capitalised?

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
```diff
@@ -13,8 +13,14 @@ object MessageToFacebook {
 
   def textMessage(id: String, message: String) =
     MessageToFacebook(Id(id),
-               Some(Message(Some(message),None,None,None)),
-               None, None)
+     Some(Message(Some(message),None,None,None)),
+     None, None)
+
+  def errorMessage(id: String) =
```
LATaylor-guardian: Can we better model the different types of messages as an ADT?

src/test/scala/com/gu/facebook_news_bot/UserStoreTest.scala
```diff
@@ -0,0 +1,56 @@
+package com.gu.facebook_news_bot
```
LATaylor-guardian: Should this package be `com.gu.facebook-news-bot`. Seems unusual to use underscores in a package name in a Scala project.

src/test/scala/com/gu/facebook_news_bot/UserStoreTest.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import akka.actor.ActorSystem
+import akka.event.Logging
+import com.amazonaws.services.dynamodbv2.model.ScalarAttributeType._
+import com.gu.facebook_news_bot.models.User
+import com.gu.facebook_news_bot.stores.UserStore
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.time.{Millis, Seconds, Span}
+
+class UserStoreTest extends org.scalatest.FunSpec with org.scalatest.Matchers with ScalaFutures {
```
LATaylor-guardian: import import org.scalatest.{FunSpec, Matchers}

src/main/scala/com/gu/facebook_news_bot/stores/UserStore.scala
```diff
@@ -1,21 +1,45 @@
...
-  private val usersTable = db.getTable(usersTableName)
+/**
+  * TODO -
+  * Make reads strongly consistent - requires scanamo change
+  */
+class UserStore(client: AmazonDynamoDBAsyncClient, usersTableName: String, logger: LoggingAdapter) {
 
-  def getUser(id: String): Option[User] = {
-    None
+  def getUser(id: String): Future[Option[User]] = {
```
LATaylor-guardian: All public functions should have return types.

src/main/scala/com/gu/facebook_news_bot/models/User.scala
tomrf1: To be consistent with the field we currently have in dynamodb, not sure if it matters

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
tomrf1: This is the only error message I was planning on sending to the user, the actual cause is logged elsewhere.

src/test/scala/com/gu/facebook_news_bot/UserStoreTest.scala
tomrf1: I wish, but java (and therefore scala) doesn't allow dashes in package names. The [suggested](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html) alternative is undescores

src/main/scala/com/gu/facebook_news_bot/models/User.scala
LATaylor-guardian: I'd revert to id if it doesn't break anything.

tomrf1: Just tested this - the case does need to be consistent, so I'll leave it capitalised

PR #27, opened 2016-10-05
Facebook service

tomrf1:
This sends messages to facebook messenger api, and requests user info from facebook graph api.


src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
```diff
@@ -62,3 +62,5 @@ object MessageFromFacebook {
   case class Postback(payload: String) extends AnyVal
 
 }
+
+case class FacebookUser(first_name: String, last_name: String, gender: String, locale: String, timezone: Double)
```
LATaylor-guardian: Do these also have to be underscored for serialization?

tomrf1: Yes, I could write something to convert to camelCase but it seems a bit pointless

PR #29, opened 2016-10-07
Capi service

tomrf1:
- Basic capi quering for headlines/most-viewed, with caching of results
- Began implementing 'headlines' request handling in `Events.scala` and `MainState.scala`, but this stuff is very unfinished. This PR was getting too big.
- Added route testing which sends a request to the `webhook` endpoint and verifies the response to facebook is correct. Uses dummy CAPI client and mock FB. The json is all loaded from file.

Other things:
- Added `"com.google.code.findbugs" % "jsr305"` to the build.sbt because the inclusion of caffeine cache lib was causing `javax.annotation.Nonnull not found` compiler errors. See [here](https://issues.scala-lang.org/browse/SI-8978) and [here](http://stackoverflow.com/questions/24263673/why-do-i-need-adding-artifact-jsr305-to-use-guava-14) for more.
- Circe cannot decode value classes (`extends AnyVal`). This was fun to debug.
- Changed `List` to `Seq` in FB model, because when circe encodes lists you get hilarious json like 

```
"::" : {
    "head": ...,
    "tail": ...
}
```


src/main/scala/com/gu/facebook_news_bot/services/Capi.scala
```diff
@@ -1,8 +1,52 @@
...
+
+  def doQuery(query: ItemQuery, getResults: (ItemResponse => Option[Seq[Content]])): Future[Seq[Content]] = {
+    CapiCache.get(query.toString).map(cached => Future.successful(cached)).getOrElse {
+      getResponse(query) map { response: ItemResponse =>
+
+        val results = getResults(response).filter(_.nonEmpty)
+          .orElse(response.results)   //Fall back on main results
+          .getOrElse(Nil)
+
+        CapiCache.put(query.toString, results)
```
LATaylor-guardian: Do you want to put the item into the cache if its Nil?

src/main/scala/com/gu/facebook_news_bot/services/Capi.scala
```diff
@@ -1,8 +1,52 @@
...
 
 class Capi(override val apiKey: String) extends GuardianContentClient(apiKey) with StrictLogging {
 
+  def getHeadlines(front: String, topic: Option[String]): Future[Seq[Content]] =
+    doQuery(basicItemQuery(front).showEditorsPicks(), _.editorsPicks)
+
+  def getMostViewed(front: String, topic: Option[String]): Future[Seq[Content]] =
+    doQuery(basicItemQuery(front).showMostViewed(), _.mostViewed)
+
+  def doQuery(query: ItemQuery, getResults: (ItemResponse => Option[Seq[Content]])): Future[Seq[Content]] = {
```
LATaylor-guardian: doQuery is only invoked within this class so can be private.

src/main/scala/com/gu/facebook_news_bot/services/Capi.scala
```diff
@@ -1,8 +1,52 @@
...
+}
+
+object CapiCache {
+
+  private val cache = Caffeine.newBuilder()
+    .expireAfterWrite(30, TimeUnit.SECONDS)
+    .build[String, Seq[Content]]()
+
+  def get(query: String): Option[Seq[Content]] = Option(cache.getIfPresent(query))
+  def put(query: String, results: Seq[Content]) = cache.put(query, results)
```
LATaylor-guardian: Return type `:Unit`

src/main/scala/com/gu/facebook_news_bot/services/Capi.scala
```diff
@@ -1,8 +1,52 @@
...
+    .tag("type/article")
+    .tag("-tone/minutebyminute")
+    .showElements("image")
+    .pageSize(25)   //Matches the number of editors-picks/most-viewed, in case they aren't available
+}
+
+object CapiCache {
+
+  private val cache = Caffeine.newBuilder()
+    .expireAfterWrite(30, TimeUnit.SECONDS)
```
LATaylor-guardian: Can we cache more aggressively? editors picks/most viewed are only updated every 5 minutes within CAPI.

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
```diff
@@ -6,30 +6,32 @@ case class Id(id: String) extends AnyVal
   * Messages sent to Messenger users
   */
 case class MessageToFacebook(recipient: Id,
-                             message: Option[MessageToFacebook.Message],
-                             sender_action: Option[String],
-                             notification_type: Option[String])
+                             message: Option[MessageToFacebook.Message]=None,
```
LATaylor-guardian: Spaces either side of `=`.

src/main/scala/com/gu/facebook_news_bot/services/Capi.scala
```diff
@@ -7,21 +7,24 @@ import com.gu.contentapi.client.GuardianContentClient
 import com.typesafe.scalalogging.StrictLogging
 import com.gu.contentapi.client.model.ItemQuery
 import com.gu.contentapi.client.model.v1.{Content, ItemResponse}
+import com.gu.facebook_news_bot.BotConfig
 
 import scala.concurrent.Future
 import scala.concurrent.ExecutionContext.Implicits.global
 
-class Capi(override val apiKey: String) extends GuardianContentClient(apiKey) with StrictLogging {
+class Capi extends StrictLogging {
```
LATaylor-guardian: Can this be an object now?

src/main/scala/com/gu/facebook_news_bot/state/Events.scala
```diff
@@ -1,17 +1,52 @@
...
   * Event functions define the user's state transition, and build any messages to be sent to user
   */
 private[state] object Events {
-  type Event = (User, MessageFromFacebook.Messaging) => Future[Result]
+  type Event = (User, MessageFromFacebook.Messaging, Capi, Facebook) => Future[Result]
 
-  def greeting: Event = (user: User, message: MessageFromFacebook.Messaging) => {
-    Future.successful((user, List(MessageToFacebook.textMessage(user.ID, "Hi!"))))
+  def greeting: Event = (user, message, capi, facebook) => {
+    Future.successful((changeState(user, "MAIN"), List(MessageToFacebook.textMessage(user.ID, "Hi!"))))
```
LATaylor-guardian: "MAIN" can be pulled out into a constant.

build.sbt
```diff
@@ -32,7 +33,9 @@ libraryDependencies ++= Seq(
   "com.amazonaws" % "aws-java-sdk-ec2" % "1.10.52",
   "com.gu" %% "scanamo" % "0.7.0",
   "com.typesafe.akka" %% "akka-http-testkit" % akkaVersion,
-  "org.scalatest" %% "scalatest" % "3.0.0" % "test"
+  "org.scalatest" %% "scalatest" % "3.0.0" % "test",
+  "com.gu" % "content-api-models-json" % "10.3" % Test, //this version needs to match the one used by content-api-client
```
LATaylor-guardian: Let's bump the version of this so that they align and then pull the version out here into a `val` so they don't differ.

src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -82,7 +80,7 @@ trait BotService extends CirceSupport with StrictLogging {
           })
         }
       case Failure(error) =>
-        logger.error(s"Error processing message $msg: $error")
+        logger.error(s"Error processing message $msg: ${error.getMessage}, ${error.getStackTrace}")
```
LATaylor-guardian: I think this can be done more simply with:

```
logger.error(s"Error processing message $msg: ${error.getMessage}", e)

```

src/test/scala/com/gu/facebook_news_bot/UserStoreTest.scala
```diff
@@ -12,8 +11,8 @@ class UserStoreTest extends FunSpec with Matchers with ScalaFutures {
   implicit val defaultPatience =
     PatienceConfig(timeout = Span(2, Seconds), interval = Span(15, Millis))
 
-  val tableName = "facebook-news-bot-dynamo"
-  LocalDynamoDB.withTableWithSecondaryIndex(tableName, "notificationTimeUTC-ID-index")('ID -> S)('notificationTimeUTC -> S,'ID -> S)
+  val tableName = "user-store-test"
```
LATaylor-guardian: As  above.

src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -28,11 +28,9 @@ trait BotService extends CirceSupport with StrictLogging {
   val capi: Capi
   val facebook: Facebook
   val stateHandler: StateHandler
+  val usersTable: String
 
-  lazy val userStore = {
-    val usersTable = BotConfig.aws.usersTable
-    new UserStore(dynamoClient, usersTable)
-  }
+  lazy val userStore = new UserStore(dynamoClient, usersTable)
```
tomrf1: Unfortunately it does because the `AmazonDynamoDBAsyncClient` might not be ready (the tests fail without it)

src/main/scala/com/gu/facebook_news_bot/services/Capi.scala
tomrf1: So I think the conclusion here is that it should cache Nil if for some reason capi isn't returning results, otherwise we risk repeatedly hitting capi with the same request. The calling code should then handle the empty `Seq[Content]` by returning an appropriate message to the client.

build.sbt
```diff
@@ -15,22 +15,27 @@ resolvers += Resolver.bintrayRepo("hseeberger", "maven")
...
-  "com.gu" %% "content-api-client" % "10.2",
+  "com.gu" %% "content-api-client" % "10.4",
   "io.circe" %% "circe-core" % CirceVersion,
   "io.circe" %% "circe-generic" % CirceVersion,
   "io.circe" %% "circe-parser" % CirceVersion,
   "com.typesafe.akka" %% "akka-actor" % akkaVersion,
   "com.typesafe.akka" %% "akka-contrib" % akkaVersion,
   "com.typesafe.akka" %% "akka-http-experimental" % akkaVersion,
+  "com.typesafe.akka" %% "akka-http-spray-json-experimental" % akkaVersion,
+  "com.github.ben-manes.caffeine" % "caffeine" % "1.3.3",
```
ben-manes: 2.3.3 please

src/main/scala/com/gu/facebook_news_bot/services/Capi.scala
```diff
@@ -1,8 +1,61 @@
...
+  private lazy val client = new GuardianContentClient(BotConfig.capi.key)
+
+  def getHeadlines(front: String, topic: Option[String]): Future[Seq[Content]] =
+    doQuery(basicItemQuery(front).showEditorsPicks(), _.editorsPicks)
+
+  def getMostViewed(front: String, topic: Option[String]): Future[Seq[Content]] =
+    doQuery(basicItemQuery(front).showMostViewed(), _.mostViewed)
+
+  private def doQuery(query: ItemQuery, getResults: (ItemResponse => Option[Seq[Content]])): Future[Seq[Content]] = {
+    CapiCache.get(query.toString).map(cached => Future.successful(cached)).getOrElse {
```
ben-manes: Please consider using loading methods to avoid racy get-compute-put calls

mchv:
> Added "com.google.code.findbugs" % "jsr305" to the build.sbt because the inclusion of caffeine cache lib was causing javax.annotation.Nonnull not found compiler errors. See here and here for more.

I am quite sure you can specify this dependency to be only at `compile` scope.

mchv:
@tomrf1 s/`compile` scope/`compile` stage/ . This is done by the `provided` scope. Anyway of you move to latest caffeine `2.3.x` you should not need it the `com.google.code.findbug` dependency.

LATaylor-guardian:
👍

PR #30, opened 2016-10-11
Main state

tomrf1:
Implements more of the `main` state, where users will spend most of their time. There are still some TODOs here. At the moment it just does headlines/most popular.

I've had a bit of a refactor. I decided the separation of `State`s from `Event`s was needlessly difficult to work with, so now everything happens inside a `State` object (e.g. `MainState`). A `State` must implement the `transition` method, which changes the user's state and builds any messages to send back to the user.


src/main/scala/com/gu/facebook_news_bot/models/User.scala
```diff
@@ -8,5 +8,8 @@ case class User(ID: String,
                 offsetHours: Double,
                 notificationTime: String,
                 notificationTimeUTC: String,
-                state: String,
-                version: Long)
+                state: Option[String] = None,
```
LATaylor-guardian: As above.

src/main/scala/com/gu/facebook_news_bot/services/Capi.scala
```diff
@@ -59,3 +59,13 @@ object CapiCache {
   def get(query: String): Option[Seq[Content]] = Option(cache.getIfPresent(query))
   def put(query: String, results: Seq[Content]): Unit = cache.put(query, results)
 }
+
+//TODO - implement topics
```
LATaylor-guardian: What are topics?

tomrf1: As with alexa, it's things that you can get headlines on. I'll have to hardcode some of the capi paths for now (e.g. US politics) until we have a better solution

src/main/scala/com/gu/facebook_news_bot/state/MainState.scala
```diff
@@ -1,27 +1,108 @@
...
-      else None
+  private sealed trait Event
+  private case class NewContentEvent(contentType: Option[ContentType], topic: Option[Topic] = None) extends Event
+  private case object MoreContentEvent extends Event
+  private case object GreetingEvent extends Event
+
+  private sealed trait ContentType { val name: String }
+  private case object HeadlinesType extends ContentType { val name = "headlines" }
+  private case object MostViewedType extends ContentType { val name = "most_viewed" }
+  private object ContentType {
```
LATaylor-guardian: This feels almost like you could benefit from using an Extractor here, see: http://danielwestheide.com/blog/2012/11/21/the-neophytes-guide-to-scala-part-1-extractors.html

src/main/scala/com/gu/facebook_news_bot/utils/FacebookMessageBuilding.scala
```diff
@@ -0,0 +1,68 @@
+package com.gu.facebook_news_bot.utils
+
+import com.gu.contentapi.client.model.v1.{Asset, Content}
+import com.gu.facebook_news_bot.BotConfig
+import com.gu.facebook_news_bot.models.MessageToFacebook
+import org.jsoup.Jsoup
+
+object FacebookMessageBuilding {
```
LATaylor-guardian: s/FacebookMessageBuilding/FacebookMessageBuilder/

src/main/scala/com/gu/facebook_news_bot/utils/FacebookMessageBuilding.scala
```diff
@@ -0,0 +1,68 @@
...
+      for {
+        data <- asset.typeData
+        width <- data.width
+      } yield width
+    }
+
+    val widestImageAsset = for {
+      elements <- content.elements
+      mainElement <- elements.find(_.relation == "main").orElse(elements.headOption)
+      asset <- mainElement.assets.foldLeft[Option[Asset]](None) { (widestAsset, thisAsset) =>
```
LATaylor-guardian: I find this quite hard to read:

```
asset <- mainElement.assets.foldLeft[Option[Asset]](None) { (widestAsset, thisAsset) =>
 +        val widest = widestAsset.flatMap(a => getWidth(a)).getOrElse(0)
 +        getWidth(thisAsset) collect {
 +          case width if width <= MaxImageWidth && width > widest => thisAsset
 +        } orElse widestAsset
 +      }
```

Maybe it can be pulled out into its own function?

src/main/scala/com/gu/facebook_news_bot/utils/FacebookMessageBuilding.scala
```diff
@@ -0,0 +1,68 @@
...
+import com.gu.facebook_news_bot.models.MessageToFacebook
+import org.jsoup.Jsoup
+
+object FacebookMessageBuilding {
+
+  val MaxImageWidth = 1000
+  val CarouselSize = 5  //Number of items in a carousel
+
+  def contentToCarousel(contentList: Seq[Content], offset: Int): Option[MessageToFacebook.Message] = {
+    val sliced = contentList.slice(offset, offset + CarouselSize)
```
LATaylor-guardian: Scala has a function called `take` which may be easier/clearer?

src/main/scala/com/gu/facebook_news_bot/utils/FacebookMessageBuilder.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  val MaxImageWidth = 1000
+  val CarouselSize = 5  //Number of items in a carousel
+
+  def contentToCarousel(contentList: Seq[Content], offset: Int): Option[MessageToFacebook.Message] = {
+    val sliced = contentList.slice(offset, offset + CarouselSize)
+    if (sliced.isEmpty) None
+    else {
+      val tiles = sliced.map { content =>
+        MessageToFacebook.Element(
```
LATaylor-guardian: Same as `Attachment`, much of this can move into `apply`.

src/main/scala/com/gu/facebook_news_bot/utils/FacebookMessageBuilder.scala
```diff
@@ -0,0 +1,66 @@
...
+          image_url = Some(getImageUrl(content)),
+          buttons = Some(List(MessageToFacebook.Button(`type` = "element_share")))
+        )
+      }
+      val attachment = MessageToFacebook.Attachment(tiles)
+
+      Some(MessageToFacebook.Message(
+        attachment = Some(attachment),
+        //TODO - topic quick_replies
+        quick_replies = Some(List(MessageToFacebook.QuickReply(
```
LATaylor-guardian: Same as above for `QuickReply.`

src/main/scala/com/gu/facebook_news_bot/utils/FacebookMessageBuilder.scala
```diff
@@ -0,0 +1,66 @@
...
+          content_type = "text",
+          title = Some("More stories"),
+          payload = Some("more")
+        )))
+      ))
+    }
+  }
+
+  // Look for widest image up to MaxImageWidth
+  private def getImageUrl(content: Content): String = {
```
LATaylor-guardian: This is only called from Element (which will now move into `apply`) so this can move onto the element companion object as a private function if you don't ever intend for it to be called from anywhere else? If you do, it probably no longer belongs to this object.

src/main/scala/com/gu/facebook_news_bot/utils/FacebookMessageBuilder.scala
```diff
@@ -0,0 +1,66 @@
...
+    val widestImageAsset = for {
+      elements <- content.elements
+      mainElement <- elements.find(_.relation == "main").orElse(elements.headOption)
+      asset <- getWidestImageAssetFromElement(mainElement)
+    } yield asset
+
+    widestImageAsset.flatMap(_.file).getOrElse(BotConfig.defaultImageUrl)
+  }
+
+  private def getWidestImageAssetFromElement(element: Element): Option[Asset] = {
```
LATaylor-guardian: Same as above as only called in `getImageUrl `.

src/main/scala/com/gu/facebook_news_bot/utils/FacebookMessageBuilding.scala
tomrf1: `take` doesn't let you choose the start index, so I think `slice` is the right method in this case

src/main/scala/com/gu/facebook_news_bot/utils/FacebookMessageBuilder.scala
tomrf1: I'm not sure I understand the motivation behind this. Objects like `Element` and `QuickReply` can be used in various different ways; the logic in here is very specific to building the content carousel. Wouldn't moving it clutter up the models definitions?

src/main/scala/com/gu/facebook_news_bot/utils/JsonHelpers.scala
```diff
@@ -0,0 +1,59 @@
...
+  def recursivelyRemoveNulls(json: Json): Json = {
+    json.arrayOrObject(
+      json,
+      (array: List[Json]) => Json.fromValues(array.map(recursivelyRemoveNulls)),
+      (obj: JsonObject) =>
+        Json.fromJsonObject(obj.toMap.foldLeft(obj) {
+          case (acc: JsonObject, (k, v)) => if (!v.isNull) acc.add(k, recursivelyRemoveNulls(v)) else acc.remove(k)
+        })
+    )
+  }
```
mchv: You should use `dropNullKeys` from the `Printer` see [circe code](https://github.com/travisbrown/circe/blob/f6de1fe56143ec01990fa796c0e3e30903677cf3/modules/core/shared/src/main/scala/io/circe/Printer.scala#L28)

PR #31, opened 2016-10-13
Menu etc

tomrf1:
Adds more functionality.

1. Subscription management - subscribe, unsubscribe, change time, change edition.
2. Handle the menu button clicks - these can be used regardless of the user's state.

Note - the difference between `buttons` and `quick_replies` is important. `quick_replies` disappear after the next message, whereas `buttons` persist. So `buttons` should never be context-specific. This isn't an issue in the old node app because it has no context


src/main/scala/com/gu/facebook_news_bot/state/BriefingTimeQuestionState.scala
```diff
@@ -0,0 +1,57 @@
...
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+
+/**
+  * When a user asks to subscribe to the morning briefings, they enter this state.
+  * It prompts them to specify one of the valid times.
+  * Upon receiving a valid time, the notification time is updated in dynamo and the user enters the MAIN state.
+  */
+case object BriefingTimeQuestionState extends State {
+  val name = "BRIEFING_TIME_QUESTION"
```
LATaylor-guardian: s/name/Name and can be private if not referenced outside this file.

src/main/scala/com/gu/facebook_news_bot/state/EditionQuestionState.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import scala.concurrent.Future
+
+/**
+  * When a user asks to update their edition they enter this state.
+  * It lists the possible editions.
+  * Upon receiving a valid edition, the 'front' field is updated in dynamo and the user enters the MAIN state.
+  */
+case object EditionQuestionState extends State {
+  val name = "EDITION_QUESTION"
```
LATaylor-guardian: As previously mentioned.

src/main/scala/com/gu/facebook_news_bot/state/EditionQuestionState.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  def transition(user: User, messaging: MessageFromFacebook.Messaging, capi: Capi, facebook: Facebook): Future[Result] = {
+    messaging.postback.map(MainState.onMenuButtonClick(user, _, capi, facebook)) getOrElse {
+      if (user.state.contains(name)) {
+        //There should be valid edition in either the text or quick_reply fields
+        val maybeEd = Seq("uk", "us", "australia", "international").find(ed => State.getUserInput(messaging).map(_.toLowerCase).exists(u => u.contains(ed)))
```
LATaylor-guardian: `Seq("uk", "us", "australia", "international")` can be pulled out of this function so that it's not created every time transition is invoked.

src/main/scala/com/gu/facebook_news_bot/state/BriefingTimeQuestionState.scala
tomrf1: The `name` fields are referenced by StandHandler.scala - it uses them to map the dynamo `state` field to a `State` object. I will capitalise it though

PR #32, opened 2016-10-24
Morning briefing

tomrf1:
This matches the morning briefing currently sent by the old node app.
A new actor object polls an SQS queue for users who should receive their morning briefings now, which is populated by the [lambda](https://github.com/guardian/facebook-news-bot-scheduler).
A carousel of editors-picks is sent. It also checks if the user's timezone has changed.

I've added the `variant` field to the User model (which is stored in dynamodb). This will be used for A/B testing later.

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/BotService.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/BotService.scala
@@ -8,6 +8,7 @@ import com.amazonaws.regions.Regions
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsyncClient
 import com.amazonaws.services.dynamodbv2.model.ConditionalCheckFailedException
 import com.gu.cm.Mode
+import com.gu.facebook_news_bot.briefing.MorningBriefingPoller
 import com.gu.facebook_news_bot.models.{MessageFromFacebook, MessageToFacebook}
 import com.gu.facebook_news_bot.services.{Capi, CapiImpl, Facebook, FacebookImpl}
 import de.heikoseeberger.akkahttpcirce.CirceSupport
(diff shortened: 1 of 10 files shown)


src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -110,5 +111,11 @@ object Bot extends App with BotService {
     }
   }
 
+  val poller = system.actorOf(Props(
+    classOf[MorningBriefingPoller],
+    userStore, capi, facebook
+  ))
+  poller ! Poll
```
LATaylor-guardian: It might be beneficial to have the Actor responsible for initiating the polling so that if the actor dies for some reason it will start polling again once it comes back up. We can also do this in its own thread, see for example: https://github.com/guardian/crier/blob/master/app/com/gu/crier/events/update/ContentPoller.scala

src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
```diff
@@ -0,0 +1,129 @@
...
+import org.joda.time.DateTime
+import org.joda.time.format.DateTimeFormat
+
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.util.{Failure, Success, Try}
+
+case object Poll
```
LATaylor-guardian: As a typical pattern, this lives inside an object with the same name as the class. This would also contain a props method which would take the necessary dependencies for the actor and return a `Props` with the class representing the actor instantiated. See previous example.

src/main/scala/com/gu/facebook_news_bot/services/SQS.scala
```diff
@@ -0,0 +1,17 @@
+package com.gu.facebook_news_bot.services
+
+import com.amazonaws.ClientConfiguration
+import com.amazonaws.regions.{Region, Regions}
+import com.amazonaws.services.sqs.AmazonSQSClient
+import com.gu.facebook_news_bot.BotConfig
+
+case class SQSMessageBody(MessageId: String, Message: String)
+
+trait SQS {
```
LATaylor-guardian: Can this be an object?

tomrf1:
@LATaylor-guardian latest commit has your suggested changes

LATaylor-guardian:
👍

PR #33, opened 2016-10-25
Topics

tomrf1:
Topics are things the user can get headlines or most popular stories for, e.g.
1. "politics"
2. "politics headlines" (same result as 1)
3. "popular politics"

A topic maps to one of the following entities in capi:
1. edition
2. section
3. edition/section
4. section/tag
5. "politics" is either 2 or 4 depending on the user's edition
(this logic could be borrowed for the alexa lambda rewrite)

I refer to both `edition` and `front` in the code. This is because I originally added it as `front` in dynamodb, so for backwards compatibility this needs to stay. Elsewhere it's called `edition`.

TODO - add suggested topics as quick replies after the carousel


src/main/scala/com/gu/facebook_news_bot/services/Capi.scala
```diff
@@ -11,22 +11,23 @@ import com.gu.facebook_news_bot.BotConfig
 
 import scala.concurrent.Future
 import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.matching.Regex
 
 trait Capi {
-  def getHeadlines(front: String, topic: Option[Topic]): Future[Seq[Content]]
+  def getHeadlines(edition: String, topic: Option[Topic]): Future[Seq[Content]]
```
tomrf1: Yeah I really regret using `front` at all, but it's in dynamodb now.
Here front and edition are the same thing. The user can have uk, us, au, international

LATaylor-guardian:
👍

PR #35, opened 2016-10-26
Add suggested topic quick_replies after the carousel

tomrf1:
This is consistent with the existing node app.

picture 2


LATaylor-guardian:
👍

PR #38, opened 2016-10-26
Fix throttling of facebook messages

tomrf1:
Before:
`message --> facebookActor`
After:
`message --> throttler --> facebookActor`

Oops.

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/services/Facebook.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/services/Facebook.scala
@@ -44,7 +44,7 @@ class FacebookImpl extends Facebook with CirceSupport with StrictLogging {
   throttler ! SetTarget(Some(facebookActor))
 
   def send(messages: List[MessageToFacebook]): Unit = {
-    messages.foreach(facebookActor ! _)
+    messages.foreach(throttler ! _)
   }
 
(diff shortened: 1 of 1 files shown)


src/main/scala/com/gu/facebook_news_bot/services/Facebook.scala
```diff
@@ -44,7 +44,7 @@ class FacebookImpl extends Facebook with CirceSupport with StrictLogging {
   throttler ! SetTarget(Some(facebookActor))
 
   def send(messages: List[MessageToFacebook]): Unit = {
-    messages.foreach(facebookActor ! _)
+    messages.foreach(throttler ! _)
```
tsop14: What does this magic `! _ `do?

tomrf1: So here `throttler` and `facebookActor` are actors.
We use the `!` method to send a message to an actor, if we don't expect (or care about) a reply. This line iterates over a list of messages to be sent to facebook, giving each one to the `throttler`

LATaylor-guardian: ! = tell, ?  = ask :)

LATaylor-guardian:
👍

PR #39, opened 2016-10-27
Verify signature in header

tomrf1:
See https://developers.facebook.com/docs/messenger-platform/webhook-reference#security for details.

FB includes an X-Hub-Signature header in its requests, which we can check against the hash of the body using the secret for the FB app.


LATaylor-guardian:
👍

PR #41, opened 2016-10-27
Log outgoing messages

tomrf1:
Log the messages being sent to facebook. I accidentally logged the incoming message, which is already being logged.


mchv:
When compiling 👍

PR #42, opened 2016-10-28
Menu quick replies, changes to the copy

tomrf1:
Quick replies are better than buttons because:
1. We can have more than 3
2. They vanish after the next message. Buttons persist, which means they can be clicked out of context later on

This change replaces the menu buttons with quick replies, and adds a new quick reply, "Suggest something". This is for telling the user that they can ask for topics. When we add more features in the future we can use this menu to tell users about them.

picture 2

picture 3


LATaylor-guardian:
👍

PR #46, opened 2016-10-28
Improve quick replies after carousel

tomrf1:
Two improvements to the set of quick replies after the carousel:
1. If user is currently viewing stories for a topic, the `more` quick reply should specify the topic, e.g. "More politics stories" instead of just "More stories"
2. Filter out the quick reply for the topic that is currently being viewed.

picture 3


tsop14:
Nice 👍

PR #47, opened 2016-10-28
Make text field optional. Handle and log rejections.

tomrf1:
The text field on the message from facebook should actually be optional. I discovered this when a user sent a message with a gif and no text, and the app didn't reject it properly. FB then refused to send any more messages until our app had handled it, so it basically stopped working in production.

Proof that it now rejects the gif properly:
picture 4


LATaylor-guardian:
👍

PR #48, opened 2016-10-31
Http fixes

tomrf1:
Over the weekend the bot failed to send/receive requests frequently, for various reasons. This branch fixes it.

1. It hit akka-http's limit on maximum number of open requests. I've raised the limit a bit in the config file, but I've also changed the code to use a connection pool and a request queue, to ensure we stay within the limit.
2. Some requests arrive as streams. Normally they don't (in akka terms, they arrive with type `HttpEntity.Strict`), which is why I never saw this problem in CODE. But sometimes in PROD the requests arrive as streams and so need to be handled correctly. It's important to consume the entire entity, even if we're not interested in the body of a particular message - [docs](http://doc.akka.io/docs/akka/2.4/scala/http/common/http-model.html#HttpEntity).
3. Handle FB response streams - as above.
4. Removed `attachment` field from the facebook model. We don't allow attachments anyway, but the message didn't conform to the model and so was causing rejections.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1,0 +1,5 @@
+akka.http {
+  host-connection-pool {
+    max-open-requests = 2048
+  }
+}
--- a/src/main/scala/com/gu/facebook_news_bot/BotConfig.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/BotConfig.scala
@@ -37,9 +37,13 @@ object BotConfig {
   }
 
   object facebook {
-    val url = getMandatoryString("facebook.url")
+    val host = getMandatoryString("facebook.host")
+    val version = getMandatoryString("facebook.version")
+    val port = getIntOrDefault("facebook.port", 443)
     val accessToken = getMandatoryString("facebook.accessToken")
     val secret = getMandatoryString("facebook.secret")
+    //Facebook must be https, but if running locally against a test service we need the option
+    val protocol = if (stage == Mode.Dev) "http" else "https"
   }
 
   object capi {
@@ -54,5 +58,8 @@ object BotConfig {
   }
   private def getMandatoryInt(name: String): Int = {
     Try(config.getInt(name)).getOrElse(sys.error(s"Error - missing mandatory config item, $name"))
+  }
+  private def getIntOrDefault(name: String, default: Int): Int = {
+    Try(config.getInt(name)).getOrElse(default)
   }
 }
--- a/src/main/scala/com/gu/facebook_news_bot/BotService.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/BotService.scala
@@ -3,10 +3,10 @@ package com.gu.facebook_news_bot
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.server.Directives._
 import akka.actor.ActorSystem
-import akka.http.scaladsl.model.{HttpResponse, StatusCodes}
-import akka.http.scaladsl.server.{MissingCookieRejection, RejectionHandler}
+import akka.http.scaladsl.model.{HttpEntity, HttpResponse, StatusCodes}
+import akka.http.scaladsl.server.{ExceptionHandler, RejectionHandler}
 import akka.stream.{ActorMaterializer, Materializer}
-import akka.http.scaladsl.unmarshalling.PredefinedFromEntityUnmarshallers
+import akka.http.scaladsl.unmarshalling.{PredefinedFromEntityUnmarshallers, Unmarshal}
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsyncClient
 import com.amazonaws.services.dynamodbv2.model.ConditionalCheckFailedException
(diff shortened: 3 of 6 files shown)


src/main/resources/application.conf
```diff
@@ -0,0 +1,5 @@
+akka.http {
+  host-connection-pool {
+    max-open-requests = 512
```
mchv: I think you can increase this safely to `2048`

src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -125,6 +151,8 @@ object Bot extends App with BotService {
   override implicit val executor = system.dispatcher
   override implicit val materializer = ActorMaterializer()
 
+  system.logConfiguration()
```
LATaylor-guardian: Is this meant to be here?

src/main/scala/com/gu/facebook_news_bot/BotConfig.scala
```diff
@@ -37,7 +37,11 @@ object BotConfig {
   }
 
   object facebook {
-    val url = getMandatoryString("facebook.url")
+    val host = getMandatoryString("facebook.host")
+    val version = getMandatoryString("facebook.version")
+    //Facebook must be https, but if running locally against a test service we need the option
+    val https = getBoolOrDefault("facebook.https", true)
```
LATaylor-guardian: Rather than passing the boolean as config, can't you use `BotConfig.stage == Mode.Dev` like you do in BotService.scala?

src/main/scala/com/gu/facebook_news_bot/services/Facebook.scala
```diff
@@ -47,11 +51,12 @@ class FacebookImpl extends Facebook with CirceSupport {
     messages.foreach(throttler ! _)
   }
 
+  val protocol = if (BotConfig.facebook.https) "https" else "http"
```
LATaylor-guardian: Can this be moved into `BotConfig`?

src/main/scala/com/gu/facebook_news_bot/BotService.scala
```diff
@@ -50,42 +60,58 @@ trait BotService extends CirceSupport with PredefinedFromEntityUnmarshallers {
...
-
-                messagings foreach { messaging =>
-                  //For now, ignore message receipts
-                  if (messaging.message.isDefined || messaging.postback.isDefined) processMessaging(messaging)
+      extractRequest { request =>
+        appLogger.debug(s"Received new webhook POST: $request")
+
+        optionalHeaderValueByName("X-Hub-Signature") { signature =>
+          //Ensure we have the full entity
+          onComplete(request.entity.toStrict(10.seconds)) {
```
LATaylor-guardian: Is 10 seconds maybe a little too high?

src/main/scala/com/gu/facebook_news_bot/BotService.scala
tomrf1: I thought it might be useful, but on second thoughts it's not really. I'll remove it

src/main/scala/com/gu/facebook_news_bot/services/Facebook.scala
```diff
@@ -63,20 +68,56 @@ class FacebookImpl extends Facebook with CirceSupport {
...
+      * Use a connection pool to limit the number of open http requests to the configured number,
+      * and a source queue to buffer requests if we exceed that limit.
+      */
+    private val pool = {
+      if (BotConfig.facebook.https) Http().cachedHostConnectionPoolHttps[Promise[HttpResponse]](host = BotConfig.facebook.host, port = BotConfig.facebook.port)
+      else Http().cachedHostConnectionPool[Promise[HttpResponse]](host = BotConfig.facebook.host, port = BotConfig.facebook.port)
+    }
+    private val queue = Source.queue[(HttpRequest, Promise[HttpResponse])](bufferSize = 1000, OverflowStrategy.dropNew)
+      .via(pool)
+      .toMat(Sink.foreach({
```
tomrf1: yes, but the other brackets aren't, i'll remove them

src/main/scala/com/gu/facebook_news_bot/BotService.scala
tomrf1: 5 seconds?

LATaylor-guardian: Yeah let's try that.

tomrf1:
@LATaylor-guardian just deploying those changes to CODE now, otherwise ready for +1?

src/main/scala/com/gu/facebook_news_bot/BotConfig.scala
```diff
@@ -55,4 +59,10 @@ object BotConfig {
   private def getMandatoryInt(name: String): Int = {
     Try(config.getInt(name)).getOrElse(sys.error(s"Error - missing mandatory config item, $name"))
   }
+  private def getIntOrDefault(name: String, default: Int): Int = {
+    Try(config.getInt(name)).getOrElse(default)
+  }
+  private def getBoolOrDefault(name: String, default: Boolean): Boolean = {
```
LATaylor-guardian: Is this still needed now?

LATaylor-guardian:
👍

PR #49, opened 2016-11-01
Handle 'start' event with existing user

tomrf1:
This fixes the case where an existing user deletes the messenger conversation, then opens a new one. When the `start` event is received for an existing user, the standard greeting + menu should be sent. This is different to the greeting for a new user.

Also made the greetings regex more flexible.

bot_start


shtukas:
:+1:

PR #50, opened 2016-11-01
Collections briefing

tomrf1:
### Existing briefing - `editors-picks`
Currently we only have one kind of morning briefing, which is a carousel built from the editors-picks for the user's edition. This is followed by some quick_replies suggesting topics, plus a "More stories" quick_reply.

picture 7

### New briefing - `collections`
This change introduces an alternative morning briefing. This is also a carousel, but the greeting message and the content in the carousel are defined by Editorial, using the Collections tool. This will be followed by the same quick_replies as the old briefing, except the "More stories" quick_reply will be replaced with "Popular stories".

picture 6

We use the front called `morning-briefing`. In here are two collections:
1. `morning-briefing` - the set of stories to be used.
2. `greeting-message` - has a single article whose headline is the greeting message to be used. This is a bit of a hack, but there's no other way to define a string in the front data.

Currently this briefing is only available to users with the UK edition selected.
The front must have been updated in the past 4 hours, otherwise it falls back on the `editors-picks` briefing. The results are cached locally for 2 mins.

### Testing
This PR adds *very* basic A/B testing, so that we can compare click-through from the two morning briefings. A new config value defines the proportion of users who should have the new briefing (default to 0.5). The urls in the briefings will contain a `variant` parameter with value of either `collections` or `editors-picks`.

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/briefing/CollectionsBriefing.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/briefing/CollectionsBriefing.scala
@@ -1,0 +1,152 @@
+package com.gu.facebook_news_bot.briefing
+
+import java.util.concurrent.TimeUnit
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model.{HttpMethods, HttpRequest}
+import akka.http.scaladsl.unmarshalling.Unmarshal
+import akka.stream.ActorMaterializer
+import com.github.benmanes.caffeine.cache.Caffeine
+import com.gu.facebook_news_bot.BotConfig
+import com.gu.facebook_news_bot.models.{Id, MessageToFacebook, User}
+import com.gu.facebook_news_bot.state.StateHandler.Result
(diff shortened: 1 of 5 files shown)


src/main/scala/com/gu/facebook_news_bot/briefing/CollectionsBriefing.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+    * Returns a morning briefing if the following conditions are met:
+    * 1. The user's ID puts them in the test group for variant B
+    * 2. The user's edition is supported
+    * 3. A collection can be retrieved from the cache or the NextGenApi
+    * 4. The collection is under OldNewsThresholdHours hours old, because old new is bad news.
+    */
+  def getBriefing(user: User): Future[Option[Result]] = {
+    if (user.ID.takeRight(1).toFloat / 10f < BotConfig.variantBProportion && EditionToCollectionName.contains(user.front)) {
```
LATaylor-guardian: I think `.toFloat` could throw an exception here so you may want to safeguard against here.

LATaylor-guardian:
👍

PR #51, opened 2016-11-02
Remove unused facebook user fields

tomrf1:
These fields come from the facebook graph api, but we only use the locale and timezone.


mchv:
👍

PR #52, opened 2016-11-02
Yml riff raff

tomrf1:
Brave new world.

1. Switch from deploy.json to riff-raff.yaml
2. Use latest sbt-riffraff-artifact
3. Fix tests, by giving default values to all config items.


LATaylor-guardian:
👍

PR #53, opened 2016-11-03
Update text

tomrf1:
1. "help" and "menu" now do the same thing - display the menu
2. Respond to "bye" and "thanks" etc
3. "news" now does the same thing as "headlines"
4. New "Give us feedback" quick_reply in the menu.

picture 10

And then:

picture 1


oilnam:
LGTM 👍

PR #54, opened 2016-11-07
Fix curated briefing caching

tomrf1:
This bug became apparent with this morning's briefings because Editorial curated their first one.
It was incorrectly caching messages with user IDs, so some users were receiving it more than once and some users got nothing at all. It could have been worse, as I had scaled up the number of instances and only a proportion of UK users were getting curated briefings. But still, OOPS.


LATaylor-guardian:
👍

PR #55, opened 2016-11-07
"Support the Guardian" quick reply

tomrf1:
The final quick reply after the carousel and in the menu now says "Support the Guardian":
picture 6

Clicking this brings up a carousel with 3 links: membership, contributions, and Kath's long read:
picture 7

I also made some small changes to help with running it locally.

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/BotConfig.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/BotConfig.scala
@@ -18,6 +18,7 @@ object BotConfig {
   val port = getMandatoryInt("port")
 
   val defaultImageUrl = getMandatoryString("defaultImageUrl")
+  val supportersImageUrl = getMandatoryString("supportersImageUrl")
 
   val campaignCode = getStringOrDefault("campaignCode", "fb_newsbot")
 
--- a/src/main/scala/com/gu/facebook_news_bot/BotService.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/BotService.scala
@@ -172,7 +172,9 @@ object Bot extends App with BotService {
     case None => appLogger.warn(s"No kinesis stream name found for logging.")
   }
 
-  val poller = system.actorOf(MorningBriefingPoller.props(userStore, capi, facebook))
+  val poller = PartialFunction.condOpt(BotConfig.stage != Mode.Dev) {
+    case true => system.actorOf(MorningBriefingPoller.props(userStore, capi, facebook))
+  }
 
   val bindingFuture = Http().bindAndHandle(routes, "0.0.0.0", BotConfig.port)
 }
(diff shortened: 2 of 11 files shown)


src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
```diff
@@ -1,6 +1,11 @@
 package com.gu.facebook_news_bot.models
 
+import scala.language.implicitConversions
+
 case class Id(id: String)
+object Id {
+  implicit def toId(id: String): Id = Id(id)
+}
```
mchv: No way you are going to merge this!

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
```diff
@@ -1,6 +1,11 @@
 package com.gu.facebook_news_bot.models
 
+import scala.language.implicitConversions
+
 case class Id(id: String)
```
mchv: `extends AnyVal` please

tomrf1: Believe me I tried extending AnyVal at the start, but it conflicts with circe. Why don't you like the implicit conversion :D

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
mchv: The reason is be explicit, rather than `magically` convert the type of a `String` to `ID`. Some people even force scala to compile their code (`-Yno-predef`) without any [Predef](http://www.scala-lang.org/api/2.11.8/index.html#scala.Predef$)

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
mchv: why do you need to model it as a case class then?

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
LATaylor-guardian: I have to commend the valiant effort to try and save four characters though! 😆

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
mchv: > Why don't you like the implicit conversion :D
 
I truly love magic ✨ as long it stay far from our code

tomrf1: Because it's facebook's model

src/main/scala/com/gu/facebook_news_bot/models/FacebookModels.scala
tomrf1: @LATaylor-guardian four characters across many places!

shtukas:
:+1:

chriswilk:
@tomrf1 love this

PR #58, opened 2016-11-08
Log referrals

tomrf1:
Facebook have [*just*](https://developers.facebook.com/docs/messenger-platform/referral-params) added support for a `ref` parameter in m.me links to messenger. This means we can log referrals from e.g. Instant Articles.

When a new user arrives, the app now checks the `postback` object for a `referral` object.


shtukas:
:+1:

PR #59, opened 2016-11-09
Append edition to variant name

tomrf1:
Only UK users get the `collections` briefing, so in order to compare click-through with the `editors-picks` briefing we need to filter out non-UK click-throughs.
We track which briefing they came from using the `variant` url param. This should not be included for non-uk users.


LATaylor-guardian:
👍

PR #61, opened 2016-11-11
Log morning briefing variants

tomrf1:
We need a precise count of how many of each morning briefing is sent out. This makes it easier to compare click-through from different morning briefings.


mchv:
👍

PR #62, opened 2016-11-11
Change name of timestamp field in logging

tomrf1:
This causes a type conflict with the default timestamp field in kibana. It is useful to have the timestamp here though.


LATaylor-guardian:
👍

PR #63, opened 2016-11-17
Report all akka queue errors properly

tomrf1:
This was a TODO that I forgot to do...
The app uses a queue for sending messages to facebook. `queue.offer` returns a result based on the `QueueOfferResult` sealed trait.
I've noticed a couple of error logs caused by the queue rejecting messages. This change will help clarify what the cause is, if it happens again.

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/services/Facebook.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/services/Facebook.scala
@@ -136,6 +136,7 @@ class FacebookImpl extends Facebook with CirceSupport {
         }
     }
 
+    private case class EnqueueException(result: QueueOfferResult) extends Throwable
     private def enqueue(strict: HttpEntity.Strict): Future[HttpResponse] = {
       appLogger.debug(s"Sending message to Facebook: $strict")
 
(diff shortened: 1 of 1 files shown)


shtukas:
:+1:

src/main/scala/com/gu/facebook_news_bot/services/Facebook.scala
```diff
@@ -149,7 +150,8 @@ class FacebookImpl extends Facebook with CirceSupport {
       queue.offer(request -> promise).flatMap {
         case QueueOfferResult.Enqueued => promise.future
         case QueueOfferResult.Failure(e) => Future.failed(e)
-        case _ => Future.failed(new RuntimeException())
+        case QueueOfferResult.Dropped => Future.failed(EnqueueException(QueueOfferResult.Dropped))
+        case QueueOfferResult.QueueClosed => Future.failed(EnqueueException(QueueOfferResult.QueueClosed))
```
LATaylor-guardian: Is this exhaustive?

shtukas: Sealed trait. I asked the same question :)

LATaylor-guardian:
👍

src/main/scala/com/gu/facebook_news_bot/services/Facebook.scala
tomrf1: yes - http://doc.akka.io/japi/akka/2.4/akka/stream/QueueOfferResult.html

PR #64, opened 2016-11-17
Find cloudformation by name rather than tags

davidfurey:
Riff-raff is about to support finding cloudformation stacks by tags.  The default behaviour for yaml configs will now be `cloudFormationStackByTags: true`.  Changing here to keep existing behaviour.


tomrf1:
@davidfurey thanks. Is there some documentation on how I should be doing this with tags?

davidfurey:
There is no documentation, but here is an example: https://github.com/guardian/mobile-apps-api/pull/857/files.

You will need to add some tags to your cloudformation stack

PR #65, opened 2016-11-22
Change sqs polling

tomrf1:
I noticed some pending morning briefing messages being dropped at the weekend. This is because it uses an in-memory queue with a limit of 1000 messages. Scaling up to two instances solved the problem, but really we should be able to run a single instance and process the SQS queue more sensibly.

This branch changes how we poll the SQS queue of users who are due to receive their briefing. Now, it only resumes polling when the current batch of 10 briefings has been sent, meaning the message queue should always be small.

I was going to add a second queue, so that we have one queue for morning briefing requests and one for higher priority requests. But since SQS only allows us to read 10 items at a time, it's not really worth it anymore.


LATaylor-guardian:
👍

PR #66, opened 2016-11-23
Send events to fb

tomrf1:
Currently the app sends event logs to kibana. Nobody likes Kibana.
FB now support custom event logging in their analytics tools. With this change, the app will send the same event logs to FB.

In the following image, I filtered the 'headlines' event by topic 'football'.
picture 1

This isn't as useful as I'd hoped, as there doesn't seem to be a way of comparing datasets, e.g. comparing the number of requests for each topic. Also, the smallest time interval is one day.

Note - the event name field for facebook analytics needs to be `_eventName`, whereas the field it's been using in kibana is `event`. The solution for now is just to include both fields in the logs, each with the same value.


shtukas:
:+1:

PR #67, opened 2016-12-14
Football transfers sign up

tomrf1:
This change allows users to sign up for football transfer window notifications.
It does not implement the notifications.

People can discover this in one of 2 ways:
1. A user is referred to the bot (e.g. from the interactive webpage). They may or may not be an existing user. There are 2 possibilities:
  a) The message contains a `postback`, which contains a `referral` field.
  b) The message's `referral` field is at the top level, not inside a `postback`.
2. An existing user finds the football transfers subscription settings while using the bot.

Notes:
- The 1st commit here refactors how postbacks and referrals are handled.
- The 1st commit also introduces a "yes or no" `State` trait, as there are now several states which expect yes/no answers.
- The new states for setting up football transfer notifications need access to a new dynamo table. The table stores which teams users are following. This table is accessed through the existing `UserStore` object, which now needs to be passed everywhere as a result. This isn't ideal, but it's the quickest solution.
- Subscription management has been reorganised to include football transfers as well as the old morning briefings.

#### Here's an example for a new user who has been referred:
picture 6

#### And an example of finding the menu from inside the bot:
picture 4


LATaylor-guardian:
@tomrf1 What a HERO for wanting Man Utd notifications in your example!

LATaylor-guardian:
Quick point: "Is there a team you would like to receive notifications for?"
My answer to this would be the team I want notifications for, not yes. @alastairjardine what do you think?

LATaylor-guardian:
@alastairjardine also interested in what you think to adding a team at a time? I would find this a little cumbersome.

alastairjardine:
Just spoken to Tom – not ambiguous, but yes, iterative. how many teams do people tend to follow?

LATaylor-guardian:
@alastairjardine Personally I would follow a league not a team for transfer news as I want to know all transfers in that league. But yes If I could only pick a team I guess I would pick the one I support.

tomrf1:
After speaking to @alastairjardine I'm going to make some changes to the flow - it can definitely be improved

tomrf1:
@LATaylor-guardian @alastairjardine I've updated the flow now. For the question "Is there another team you’re interested in?", it is valid for a user to answer with both:
1. yes/no,
2. the name of a team.

src/main/scala/com/gu/facebook_news_bot/state/EditionQuestionState.scala
```diff
@@ -21,16 +22,14 @@ case object EditionQuestionState extends State {
 
   private case class EditionEvent(id: String, event: String = "change_edition", _eventName: String = "change_edition", edition: String) extends LogEvent
 
-  def transition(user: User, messaging: MessageFromFacebook.Messaging, capi: Capi, facebook: Facebook): Future[Result] = {
-    messaging.postback.map(MainState.onMenuButtonClick(user, _, capi, facebook)) getOrElse {
```
LATaylor-guardian: And here?

src/main/scala/com/gu/facebook_news_bot/state/YesOrNoState.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import com.gu.facebook_news_bot.models.{MessageFromFacebook, MessageToFacebook, User}
+import com.gu.facebook_news_bot.services.{Capi, Facebook}
+import com.gu.facebook_news_bot.state.StateHandler.Result
+import com.gu.facebook_news_bot.stores.UserStore
+
+import scala.concurrent.Future
+
+trait YesOrNoState extends State {
+  private val YesPattern = """\b(yes|yeah|yep|sure|ok|okay)\b""".r.unanchored
```
LATaylor-guardian: This is also declared in `EnterTeamsState`

src/main/scala/com/gu/facebook_news_bot/state/YesOrNoState.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  def transition(user: User, messaging: MessageFromFacebook.Messaging, capi: Capi, facebook: Facebook, store: UserStore): Future[Result] = {
+    if (State.getUserInput(messaging).exists(s => YesPattern.findFirstIn(s.toLowerCase).isDefined)) yes(user, facebook)
+    else no(user) //if they haven't said yes then it's probably a no
```
LATaylor-guardian: This comment is unnecessary.

LATaylor-guardian:
This looks good to me 👍

PR #71, opened 2016-12-20
Store transfer rumour notification times

tomrf1:
As well as transfer notifications by team, users will get a daily link to the "Rumour Mill" series. This will go out at midday in the user's timezone. This time will be stored in a new field in the dynamodb users table - `footballRumoursTimeUTC`.

A lambda will check for users who are due to receive their notification using this field.


shtukas:
:+1: Look good to me!

PR #72, opened 2016-12-21
Do not rely on the scheduler's User model

tomrf1:
The lambda that schedules morning briefings publishes the user data from the dynamo table to an SNS topic, to be consumed by the bot app.
This was intended to avoid doing an extra dynamo read in the bot app. But it's a really bad idea because it means the User model in the lambda has to be the same as the User model in the bot app.

This change makes the bot app do the dynamo lookup as well.


LATaylor-guardian:
👍

PR #73, opened 2016-12-21
Support 'free agent' transfers

tomrf1:
The Sports desk has asked us to handle the case where a player with no previous team joins a new team.

@tsop14


shtukas:
:+1:

LATaylor-guardian:
The "Joey Barton" effect!

PR #75, opened 2016-12-22
Advertise football transfers notifications

tomrf1:
On the 1st January ask all users if they'd like to subscribe to football transfers notifications immediately after their morning briefing (if they're not already subscribed).

Also in this PR:
- make `Facebook.send(messages)` send the next message only after receiving a response from FB. This doesn't necessarily guarantee delivery order, but it seems reliable so far.
- log when people say "no" to the subscription question.

picture 3


LATaylor-guardian:
👍

PR #77, opened 2016-12-23
Log whether a user is subscribed to morning briefing when they respond to the football transfers question

tomrf1:
We'll hoping to refer lots of new users from football-related content on-platform, but it will be useful to know how many existing subscribers are interested in football transfers.

`notificationTimeUTC == "-"` means a user is not subscribed to the morning briefing.


shtukas:
:+1:

PR #78, opened 2016-12-23
Ask users if they want morning briefing after transfers question

tomrf1:
And some copy changes.
@chriswilk 

picture 5


LATaylor-guardian:
👍

PR #80, opened 2017-01-03
Handle SQS message deletion failures

tomrf1:
In late December, 9 people received the transfer rumours link twice in a row.
This happened because it failed to delete the messages from the SQS queue on the 1st attempt, meaning the same 9 messages were read again when the app next polled the queue.

_It is always preferable to send no message than to risk sending it multiple times._

With this change, the app defends against message deletion failures. Each `SQSPoller` actor (we currently have 3, for different kinds of notifications) will try to delete a batch of messages from SQS 3 more times before giving up. If it still can't delete them, the actor kills itself (the supervisor does not restart the actor in this case) to avoid spamming users. This means the scheduled messages on that queue will no longer be sent to users until the app is restarted, but the bot will still be contactable.

We already have cloudwatch alarms monitoring for when the SQS messages are not being read.

Tested in CODE by revoking the sqs delete permission.

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/utils/SQSPoller.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/utils/SQSPoller.scala
@@ -2,7 +2,7 @@ package com.gu.facebook_news_bot.utils
 
 import java.util.concurrent.Executors
 
-import akka.actor.Actor
+import akka.actor.{Actor, Kill}
 import com.amazonaws.services.sqs.model.{DeleteMessageBatchRequest, DeleteMessageBatchRequestEntry, Message, ReceiveMessageRequest}
 import com.google.common.util.concurrent.ThreadFactoryBuilder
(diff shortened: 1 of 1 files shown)


mchv:
The code looks fine, but I would like to discuss the strategy. 
Rather than killing the actor we could for instance add a [circuit breaker](http://doc.akka.io/docs/akka/snapshot/common/circuitbreaker.html) that will allow the app to recover itself after a deletion are not failing anymore

LATaylor-guardian:
Worth linking to the docs here for how the supervisor handles Kill: http://doc.akka.io/docs/akka/snapshot/scala/fault-tolerance.html

mchv:
👍 as you can merge the current code, and improve in a future request

tomrf1:
@mchv good idea. For now we can act after the cloudwatch alarm (if it ever happens)

PR #81, opened 2017-01-03
Use fromClub crest if no toClub crest available

tomrf1:
Currently the app uses the crest of the club the player is going to, but if we don't have a crest for that club then it falls back on a default image.

This happens a lot, e.g. a player going out on loan to a smaller club. I think it's better to fall back on the crest of the club they're coming from.

E.g.
picture 10


shtukas:
:+1:

LATaylor-guardian:
I agree 👍

PR #82, opened 2017-01-03
Change subscription settings flow

tomrf1:
1. Ask which edition a user wants while subscribing to the morning briefing. Previously the edition could only be set from the subscriptions menu, which isn't obvious.
picture 11

2. Handle unsubscribe requests better. Previously saying "unsubscribe" out of context would unsubscribe from the morning briefing. Now it asks the user which service to unsubscribe from, based on their current subscriptions.
picture 12


alastairjardine:
"All" or would "Both" be more appropriate/linguistically correct?

tomrf1:
Perhaps, though the idea is that we may add more subscription types in the future

shtukas:
:+1:

PR #83, opened 2017-01-04
Remove logic for football transfers question on new year

tomrf1:
Also rewrite the horrible isNewYear function I wrote on new years eve


LATaylor-guardian:
👍

PR #84, opened 2017-01-04
Greet existing user on referral

tomrf1:
E.g. when an existing user clicks the Messenger link in an Instant Article.
Currently it says "Sorry, I didn't understand that" etc.
It should just greet them.


LATaylor-guardian:
👍

PR #85, opened 2017-01-05
Handle failed scheduled messages

tomrf1:
The actor responsible for queueing up messages to be sent to facebook sometimes fails to respond (akka.pattern.AskTimeoutException) while processing many scheduled messages.
This has been causing the actor responsible for preparing football transfer rumour messages to stop working.

This change ensures that the `SQSPoller` actors continue polling after the facebook actor has failed to respond.

I will investigate the cause of the timeouts separately.


shtukas:
:+1:

LATaylor-guardian:
👍

PR #86, opened 2017-01-05
Fix handling of 'released' players

tomrf1:
The field in the spreadsheet is called `released`.
Also, Editorial _have_ asked for "has" to be changed to "have".

picture 13


mchv:
👍 💯  for the screenshort

PR #88, opened 2017-01-06
Increase FB actor timeout

tomrf1:
Messages to be sent to facebook are sent to an actor, which maintains a queue of pending http requests.
When a message is given to this actor, the result is expected some time in the future. Recently, the ask request to the actor has occasionally timed out while processing many morning-briefings and transfer rumours simultaneously. The current actor timout of 5 seconds clearly isn't enough at these times.


shtukas:
:+1:

PR #89, opened 2017-01-09
Improve recording of uncontactable users

tomrf1:
We record when users become uncontactable in order to get a true subscriber count.

If a user does not unusubscribe from notifications formally through our bot, they can prevent us from sending messages in one of the followings ways:
1. Delete the conversation - this is the most common. This means we cannot get the user's data from the FB graph api, and we cannot send them messages. We already record this case in dynamodb.
2. Click "Turn off all messages" in the Manage Messages menu. This means we can still get the user's data from FB graph api, but cannot send them messages. We do not currently record this case.

This change does the following:
1. Records when users become uncontactable through the 2nd option above.
2. Moves the notifications logic in `MorningBriefingPoller` to the new `Notifier` object, for use by other types of notifications. Transfer rumours notifications now uses it.

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/football_transfers/FootballTransferRumoursPoller.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/football_transfers/FootballTransferRumoursPoller.scala
@@ -6,13 +6,14 @@ import akka.actor.Props
 import com.github.benmanes.caffeine.cache.Caffeine
 import com.gu.contentapi.client.model.v1.Content
 import com.gu.facebook_news_bot.BotConfig
-import com.gu.facebook_news_bot.models.{Id, MessageToFacebook}
+import com.gu.facebook_news_bot.models.{Id, MessageToFacebook, User}
 import com.gu.facebook_news_bot.services.Facebook.FacebookMessageResult
 import com.gu.facebook_news_bot.services.{Capi, Facebook, FacebookEvents, SQSMessageBody}
 import com.gu.facebook_news_bot.utils.Loggers.LogEvent
-import com.gu.facebook_news_bot.utils.{FacebookMessageBuilder, JsonHelpers, SQSPoller}
+import com.gu.facebook_news_bot.utils.{FacebookMessageBuilder, JsonHelpers, Notifier, SQSPoller}
 import com.gu.facebook_news_bot.utils.Loggers._
 import com.gu.facebook_news_bot.football_transfers.FootballTransferRumoursPoller._
+import com.gu.facebook_news_bot.stores.UserStore
 import org.jsoup.Jsoup
 import io.circe.generic.auto._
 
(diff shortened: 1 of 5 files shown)


src/main/scala/com/gu/facebook_news_bot/football_transfers/FootballTransferRumoursPoller.scala
```diff
@@ -25,16 +29,27 @@ import scala.concurrent.duration._
...
+class FootballTransferRumoursPoller(val facebook: Facebook, val capi: Capi, val userStore: UserStore) extends SQSPoller {
   val SQSName = BotConfig.football.rumoursSQSName
   override val PollPeriod = 2000.millis
 
   protected def process(messageBody: SQSMessageBody): Future[List[FacebookMessageResult]] = {
-    JsonHelpers.decodeJson[UserFootballTransferRumour](messageBody.Message).map(processRumour) getOrElse Future.successful(Nil)
+    JsonHelpers.decodeJson[UserFootballTransferRumour](messageBody.Message).map { rumour =>
+
+      val futureMaybeUserMessages: OptionT[Future, (User, List[MessageToFacebook])] = for {
+        user <- OptionT(Notifier.getUser(rumour.userId, facebook, userStore))
```
LATaylor-guardian: Nice use of monad transformers.

src/main/scala/com/gu/facebook_news_bot/football_transfers/FootballTransferRumoursPoller.scala
```diff
@@ -25,16 +29,27 @@ import scala.concurrent.duration._
...
+      futureMaybeUserMessages.value.flatMap {
+        case Some((user, messages)) => Notifier.sendAndUpdate(user, messages, facebook, userStore)
+        case None => Future.successful(Nil)
+      }
+    }.getOrElse(Future.successful(Nil))
   }
 
-  private def processRumour(rumour: UserFootballTransferRumour): Future[List[FacebookMessageResult]] = {
-    val futureMaybeMessage: Future[Option[MessageToFacebook.Message]] = FootballTransferRumoursCache.get(rumour.articleId).map(m => Future.successful(Some(m)))
+  def getMessages(rumour: UserFootballTransferRumour): Future[Option[List[MessageToFacebook]]] = {
```
LATaylor-guardian: Can `Future[Option[List[MessageToFacebook]]]` not be made simpler by mapping over the future and doing a `.getOrElse Nil` on the option. This would reduce the type signature to `Future[List[MessageToFacebook]]`.

src/main/scala/com/gu/facebook_news_bot/utils/Notifier.scala
```diff
@@ -0,0 +1,136 @@
...
+    } else {
+      updateUser(user, userStore)
+      Future.successful(Nil)
+    }
+  }
+
+  private def changeNotificationTime(user: User, timezone: Double): User = {
+    val newNotificationTimeUTC: String = {
+      if (user.notificationTime != "-") {
+        val notifyTime = DateTime.parse(user.notificationTime, DateTimeFormat.forPattern("HH:mm"))
```
LATaylor-guardian: You might want to extract `DateTimeFormat.forPattern("HH:mm")` out as an object member so it's only created once.

src/main/scala/com/gu/facebook_news_bot/football_transfers/FootballTransferRumoursPoller.scala
tomrf1: I agree that returning `Option[List[...]]` is a bit silly. I've just pushed a commit to return Nil instead. This means losing the monad transformer, as we're no longer getting a pair of `Future[Option[_]]`.
I'm undecided about which approach looks nicer now, what do you think?

LATaylor-guardian: I think it's fine now. It's certainly no less readable and you've reduced the complexity so I'd say that's a win.

LATaylor-guardian:
👍

PR #95, opened 2017-01-24
Change in cloudformation: DynamoDB table for Oscar winner predictions

shtukas:


tomrf1:
Why do you need indexes on each category?

shtukas:
We actually don't. Let me update that.

@annebyrne

shtukas:
@tomrf1 @annebyrne Updated! Also while being at it, I have renamed `bestFilm` into `bestPicture`.

tomrf1:
Great. And now that the `best*` fields are no longer used in indexes, you can remove them from the cloudformation completely. I think your template will actually be rejected if you leave them in.
The app will decide the names/types of these fields.

shtukas:
Oh. That's useful to know. Give me one sec.

shtukas:
Ok done, only `ID` is required.

tomrf1:
👍

PR #100, opened 2017-01-25
Support custom handling of postbacks

tomrf1:
In general, the receipt of a `postback` means the user has clicked a button.
It's best to avoid using buttons for behaviour that is specific to a state, as buttons persist after the conversation has moved on, meaning the user can later click a button out of context.

However, for the oscar nominations feature we have a legitimate reason to use buttons - for choosing predictions from a carousel.

So this change makes it possible for a state to override handling of postbacks. By default postbacks will still be handled by the `MAIN` state.


shtukas:
LGTM :+1:

PR #102, opened 2017-01-26
Model for Academy Award Nominees

annebyrne:


shtukas:
:+1:

shtukas:
:+1:

tomrf1:
👍

src/main/scala/com/gu/facebook_news_bot/models/UserNoms.scala
```diff
@@ -5,3 +5,49 @@ case class UserNoms(ID: String,
                     bestDirector: Option[String] = None,
                     bestActor: Option[String] = None,
                     bestActress: Option[String] = None)
+
+case class IndividualNominee(name: String, pictureUrl: String)
```
LATaylor-guardian: `case class IndividualNominee(name: String, pictureUrl: String = "https://media.guim.co.uk/82ef416b00febc774665554973c2965c7f82b819/0_191_2266_1360/140.jpgq")`

Then you don't need to write defaultPictureUrl everywhere where you are instantiating `IndividualNominee` :)

PR #103, opened 2017-01-30
Decoder for messages seq

tomrf1:
For loading in a JSON array of expected messages for the tests.

### DIFF
--- a/src/test/scala/com/gu/facebook_news_bot/util/JsonHelpers.scala
+++ b/src/test/scala/com/gu/facebook_news_bot/util/JsonHelpers.scala
@@ -1,11 +1,21 @@
 package com.gu.facebook_news_bot.util
 
-import io.circe.{Decoder, Json}
+import cats.data.Xor
+import com.gu.facebook_news_bot.models.MessageToFacebook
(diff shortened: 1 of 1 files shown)


shtukas:
:+1:

src/test/scala/com/gu/facebook_news_bot/util/JsonHelpers.scala
```diff
@@ -1,11 +1,21 @@
...
 
 import scala.io.Source._
 import io.circe.parser._
+import io.circe.generic.auto._
 
 object JsonHelpers {
+
+  implicit def messagesDecoder = Decoder.instance[Seq[MessageToFacebook]] { c =>
+    c.top.asArray.map { array: List[Json] =>
+      Xor.Right(array.flatMap(_.as[MessageToFacebook].toOption))
```
LATaylor-guardian: It's probably worth upgrading Circe and Cats as some point soon so that we no longer need to use Xor.

tomrf1: agreed!

PR #104, opened 2017-01-30
Updates in OscarsNomsStates and some tests

shtukas:
Work done with @annebyrne last week. I squashed all our commits.

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
@@ -1,0 +1,205 @@
+package com.gu.facebook_news_bot.state
+
+import com.gu.facebook_news_bot.models._
+import com.gu.facebook_news_bot.services.{Capi, Facebook}
+import com.gu.facebook_news_bot.state.StateHandler.Result
+import com.gu.facebook_news_bot.stores.UserStore
+import com.gu.facebook_news_bot.utils.JsonHelpers
+import com.gu.facebook_news_bot.utils.Loggers.LogEvent
+import io.circe.generic.auto._
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+sealed trait NominationCategory { val name: String }
+case object BestPicture extends NominationCategory { val name = "Best Picture" }
+case object BestDirector extends NominationCategory { val name = "Best Director" }
+case object BestActress extends NominationCategory { val name = "Best Actress" }
+case object BestActor extends NominationCategory { val name = "Best Actor" }
+
+object OscarsNomsStates {
+
+  case object InitialQuestionState extends YesOrNoState {
+
+    val Name = "OSCARS_NOMS_INITIAL_QUESTION"
+
+    val Question = "Welcome to the Guardian Academy. Choose your favourite Oscar contenders and we'll let you know how your picks do on the night."
+
+    protected def getQuestionText(user: User) = {
+      if (user.version.contains(0)) s"Hi, I'm the Guardian chatbot. $Question"
+      else Question
+    }
+
+    protected def yes(user: User, facebook: Facebook): Future[Result] = EnterNomsState.question(user)
+
+    private case class NoEvent(id: String, event: String = "oscars_noms_subscribe_no", _eventName: String = "oscars_noms_subscribe_no") extends LogEvent
+
+    protected def no(user: User): Future[Result] = {
+      State.log(NoEvent(id = user.ID))
+      MainState.menu(user, "Ok. Is there anything else I can help you with?")
+    }
+
+  }
+
+  case object EnterNomsState extends State {
+
+    val Name = "OSCARS_ENTER_NOMS"
+
+    case class Prediction(name: String, category: String)
+
+    def transition(user: User, messaging: MessageFromFacebook.Messaging, capi: Capi, facebook: Facebook, store: UserStore): Future[Result] = {
+      State.getUserInput(messaging) match {
+
+        case Some(text) => store.OscarsStore.getUserNominations(user.ID).map { maybeUserNoms =>
+          val userNoms = maybeUserNoms.getOrElse(UserNoms(user.ID))
+          categoryAwaitingPrediction(userNoms) match {
+            case Some(category) =>
+              val nominees = buildNominationCarousel(category, user)
+              val message = MessageToFacebook.textMessage(user.ID, s"Sorry I didn't get that. Who do you think will win ${category.name}?")
+              (user, List(message, nominees))
+            case None => finishedPlaying(user)
+          }
+        }
+        case None => MainState.menu(user, "Ok. Is there anything else I can help you with?")
+      }
+    }
+
+    def question(user: User, text: Option[String] = None): Future[Result] = {
+      requestBestPicture(user)
+    }
+
+    private def requestBestPicture(user: User): Future[Result] = {
+      val message = MessageToFacebook.textMessage(user.ID, "Great. Let’s start with Best Picture. Which of these deserves the Oscar?")
+      val categoryNominees = buildNominationCarousel(BestPicture, user)
+      Future.successful(State.changeState(user, Name), List(message,categoryNominees))
+    }
+    
+    override def onPostback(user: User, postback: MessageFromFacebook.Postback, capi: Capi, facebook: Facebook, store: UserStore): Future[Result] = {
+      val userChoice = postback.payload
+      JsonHelpers.decodeJson[Prediction](userChoice) match {
+        case Some(prediction) => processPrediction(prediction, user, store)
+        case None =>
+          val message = MessageToFacebook.textMessage(user.ID, "Sorry I didn't catch that. Could you try again?")
+          Future.successful(State.changeState(user, Name), List(message))
+      }
+    }
+
+    private def processPrediction(prediction: Prediction, user: User, store: UserStore): Future[Result] = {
+      store.OscarsStore.getUserNominations(user.ID).map { maybeUserNoms =>
+        val userNoms = maybeUserNoms.getOrElse(UserNoms(user.ID))
+
+        getCategory(prediction.category) match {
+          case Some(category) =>
+            if (categoryFieldIsEmpty(category, userNoms)) {
+              val newUserNoms = updateUserNoms(userNoms, category, prediction.name)
+              store.OscarsStore.putUserNominations(newUserNoms)
+              requestFollowUpPrediction(user, newUserNoms, prediction)
+            } else {
+              val expectedCategory = categoryAwaitingPrediction(userNoms)
+              expectedCategory match {
+                case Some(category) => {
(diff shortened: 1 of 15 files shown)


src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -0,0 +1,166 @@
...
+
+}
+
+object OscarsNomsStates {
+
+  case object InitialQuestionState extends YesOrNoState {
+
+    val Name = "OSCARS_NOMS_INITIAL_QUESTION"
+
+    val Question = "Welcome to the Guardian Academy. Choose your favorite Oscar contenders and we’ll let you know how your picks do on the night, how your taste compares to other readers and we’ll keep you updated with the latest news ahead of the awards. Ready to vote?"
```
tomrf1: Is this the final copy? What is the Guardian Academy?
Also you spelled 'favourite' wrong :)

annebyrne: @tomrf1 Yup this was the copy Nathan shared with us. Good spot on typo!

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -0,0 +1,166 @@
...
+trait NominationCategory
+object BestPicture extends NominationCategory
+object BestDirector extends NominationCategory
+object BestActress extends NominationCategory
+object BestActor extends NominationCategory
+
+object OscarsNomsStatesHelper {
+
+  def missingCategoryFromUserNominations(userNoms : UserNoms): NominationCategory = {
+    if(userNoms.bestPicture.isEmpty){
```
tomrf1: Just a style thing - I like to have spaces before/after brackets and braces.
Also, I think this is a case where it looks nicer without the braces:
```
if (userNoms.bestPicture.isEmpty) BestPicture
else if (userNoms.bestDirector.isEmpty) BestDirector
...
```

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -0,0 +1,166 @@
...
+import com.gu.facebook_news_bot.services.{Capi, Facebook}
+import com.gu.facebook_news_bot.state.StateHandler.Result
+import com.gu.facebook_news_bot.stores.UserStore
+import com.gu.facebook_news_bot.utils.Loggers.LogEvent
+import io.circe.generic.auto._
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+trait NominationCategory
```
LATaylor-guardian: Should this not be a sealed trait?

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -0,0 +1,166 @@
...
+object BestPicture extends NominationCategory
```
LATaylor-guardian: Similarly should this not be a case object?

src/test/scala/com/gu/facebook_news_bot/OscarsNomsTest.scala
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  it("should ask a new user if they want to submit their Oscar winners predictions") {
+    routeTest(
+      "src/test/resources/facebookRequests/oscarsNoms/newUser.json",
+      "src/test/resources/facebookResponses/oscarsNoms/newUser.json"
+    )
+  }
+
+//  it("should ask a user how else they can help if user does not want to play") {
```
tomrf1: what is this test for?

src/test/scala/com/gu/facebook_news_bot/OscarsNomsTest.scala
```diff
@@ -0,0 +1,60 @@
...
+//  }
+
+  it("should ask a user who they think will win the Best Picture category") {
+    routeTest(
+      "src/test/resources/facebookRequests/oscarsNoms/enterBestPicture.json",
+      "src/test/resources/facebookResponses/oscarsNoms/enterBestPicture.json"
+    )
+  }
+
+  /*it("should confirm a user's submission for Best Picture and ask who they think will win Best Director") {
```
tomrf1: worth adding this test

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -0,0 +1,166 @@
...
+object BestDirector extends NominationCategory
+object BestActress extends NominationCategory
+object BestActor extends NominationCategory
+
+object OscarsNomsStatesHelper {
```
LATaylor-guardian: Not really a fan of `helper` objects. Can we provide a more meaningful name or group the functions in this object in a different way?

src/test/scala/com/gu/facebook_news_bot/OscarsNomsTest.scala
LATaylor-guardian: These can be deleted if they are not being used.

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -0,0 +1,166 @@
...
+      val futureMaybeExistingUserNominations = store.OscarsStore.getUserNominations(user.ID)
+      futureMaybeExistingUserNominations.flatMap{ maybeExistinguserNominations =>
+        val existingUserNominations =  maybeExistinguserNominations.getOrElse(UserNoms(user.ID))
+        val newUserNominations = OscarsNomsStatesHelper.updateUserNominations(existingUserNominations, userChoice)
+        store.OscarsStore.putUserNominations(newUserNominations)
+        requestFollowUpPrediction(user, Some(newUserNominations))
+      }
+    }
+
+    def requestFollowUpPrediction(user: User, maybeUserNoms: Option[UserNoms]): Future[Result] = {
```
tomrf1: Feels wrong to be passing in an `Option[UserNoms]` - can we make `transition` create one first?

src/test/scala/com/gu/facebook_news_bot/OscarsNomsTest.scala
LATaylor-guardian: As above.

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
tomrf1: They can probably just go in `EnterNomsState`, since that's where they're used.

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -116,6 +42,67 @@ object OscarsNomsStates {
 
   case object EnterNomsState extends State {
 
+    def missingCategoryFromUserNominations(userNoms : UserNoms): NominationCategory = {
```
LATaylor-guardian: If the functions you moved into here are only called within this object then they may be private.

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -0,0 +1,155 @@
...
+            val userNoms = maybeUserNominations.getOrElse(UserNoms(user.ID))
+            requestFollowUpPrediction(user, userNoms)
+          })
+        // The above expression means that as it stands the user is forced to finish the game
+        case None => notPlaying(user)
+      }
+    }
+
+    private def missingCategoryFromUserNominations(userNoms : UserNoms): NominationCategory = {
+      if( userNoms.bestPicture.isEmpty ) BestPicture
```
LATaylor-guardian: There's no need to change it, but when @tomrf1 mentioned a space he meant between the if and the first bracket. So `if (true = true)` rather than `if( true = true)`.

LATaylor-guardian:
Well done :) Looks good from my point of view now.

src/test/resources/facebookRequests/oscarsNoms/userDecline.json
```diff
@@ -0,0 +1,24 @@
+{
```
tomrf1: These test files arent being used. Is it worth adding a test for them?

annebyrne: Good spot! Will add in now.

tomrf1:
👍 assuming it works as expected in CODE.
Then we just need the 'update type' question and the share button and we're done?

annebyrne:
Refactored to remove helper methods & make bot work in CODE. @tomrf1 @LATaylor-guardian

shtukas:
I just squashed everything. :+1:

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -0,0 +1,208 @@
...
+
+  case object EnterNomsState extends State {
+
+    val Name = "OSCARS_ENTER_NOMS"
+
+    case class Prediction(name: String, category: String)
+
+    def transition(user: User, messaging: MessageFromFacebook.Messaging, capi: Capi, facebook: Facebook, store: UserStore): Future[Result] = {
+      State.getUserInput(messaging) match {
+        case Some(text) => store.OscarsStore.getUserNominations(user.ID).map(
```
tomrf1: Just a style thing but I find the following more readable (also avoids using the name `category` twice):
```
case Some(text) => store.OscarsStore.getUserNominations(user.ID).map { maybeUserNoms =>
  val userNoms = maybeUserNoms.getOrElse(UserNoms(user.ID))

  categoryAwaitingPrediction(userNoms) match {
    case Some(category) =>
      val nominees = buildNominationCarousel(category, user)
      val message = MessageToFacebook.textMessage(user.ID, s"Sorry I didn't get that. Who do you think will win ${category.name}?")
      (user, List(message, nominees))

    case None => finishedPlaying(user)
  }
}
```

src/main/scala/com/gu/facebook_news_bot/state/OscarsNomsStates.scala
```diff
@@ -0,0 +1,208 @@
...
+          maybeUserNoms => {
+            val userNoms = maybeUserNoms.getOrElse(UserNoms(user.ID))
+            val category = categoryAwaitingPrediction(userNoms)
+
+            category match {
+              case Some(category) => {
+                val nominees = buildNominationCarousel(category, user)
+                val message = MessageToFacebook.textMessage(user.ID, s"Sorry I didn't get that. Who do you think will win ${category.name}?")
+                (State.changeState(user, Name), List(message, nominees))
```
tomrf1: No need to update `user` here

tomrf1:
👍

PR #105, opened 2017-01-31
Allow read / write access to Oscars DB table

annebyrne:


tomrf1:
Looks fine, let's update the CODE stack and redeploy

PR #106, opened 2017-02-03
Football transfers feedback

tomrf1:
We will send out a one-off message to all football transfers subscribers, to ask for feedback.

- Made `FeedbackState` a trait so that it can be used for feedback on specific things.
- Added a new poller to process these notifications. This uses the existing SQS for the football rumours, as we need to send these out today.

TODO - build a more general way of sending out mass messages to a subset of users like this.


shtukas:
LGTM :+1:, but note that I am not familiar with the logic inside `FootballTransfersFeedbackPoller`

PR #109, opened 2017-02-06
Custom briefing pub

tomrf1:
Support the publication of custom morning briefings. The user can choose 1-2 topics to include in the carousel. The 5 stories in the carousel are picked as follows:
1. editors-picks[0]
2. topic1[0]
3. editors-picks[1]
4. topic2[0] || topic1[1]
5. editors-picks[2]

The urls will end with `&variant=custom-briefing-<edition>`


LATaylor-guardian:
👍

PR #111, opened 2017-02-06
Send readers' editor url for long messages

tomrf1:
We get quite a lot of editorial-related comments via the feedback mechanism in the bot.
This change makes it a bit clearer that it's looking for feedback on the bot, and sends a link to Readers' Editor contact details for general messages that are over 200 chars.

picture 5


tsop14:
👍 

url is `.../2013/...` ☹️

PR #112, opened 2017-02-07
Advertise custom briefings

tomrf1:
On 8th Feb all morning briefings will be followed by a message asking if they'd like to customise their briefings.

picture 6

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
@@ -6,13 +6,14 @@ import com.gu.facebook_news_bot.briefing.MorningBriefingPoller._
 import com.gu.facebook_news_bot.models.{MessageToFacebook, User}
 import com.gu.facebook_news_bot.services.Facebook.FacebookMessageResult
 import com.gu.facebook_news_bot.services.{Capi, Facebook, FacebookEvents, SQSMessageBody}
-import com.gu.facebook_news_bot.state.MainState
+import com.gu.facebook_news_bot.state.{CustomBriefingQuestionState, MainState}
 import com.gu.facebook_news_bot.state.StateHandler._
 import com.gu.facebook_news_bot.stores.UserStore
 import com.gu.facebook_news_bot.utils.Loggers._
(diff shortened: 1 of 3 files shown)


mchv:
👍  Interested to see the results!

src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
```diff
@@ -73,8 +76,16 @@ class MorningBriefingPoller(val userStore: UserStore, val capi: Capi, val facebo
       val variant = s"editors-picks-${user.front}"
       logBriefing(user.ID, variant)
 
-      MainState.getHeadlines(user, capi, Some(variant)) map { case (updatedUser, messages) =>
-        (updatedUser, morningMessage(updatedUser) :: messages)
+      //TODO - remove later
+      if (user.briefingTopic1.isEmpty && SQSPoller.isDate(user.offsetHours, advertiseCustomBriefingDate)) {
+        for {
+          (_, headlinesMessages) <- MainState.getHeadlines(user, capi, Some(variant))
+          (updatedUser, transfersMessages) <- CustomBriefingQuestionState.question(user, Some("I can now personalise your briefing. Do you want to choose a favourite topic?"))
```
tsop14: In your example you replied with `football` but how would the user know this will be accepted - should there be some sort of clue to accepted responses?

tsop14: What happens if they reply with `yes`? Which would be a valid exchange.

tomrf1: If you say "yes" then it replies "Ok, which topic?" followed by the list of topics again.
If the topic is accepted then a confirmation is sent (text depends on whether it was the 1st or 2nd topic chosen).
If the topic is unknown then it says "Sorry, I don't know that topic. Are you interested in any of these?"

tomrf1: The initial question is followed by a set of quick replies, just realised that isn't obvious from the screenshot!
picture 7

tsop14: Nice!

PR #113, opened 2017-02-07
Add images to Oscar Nominations Game

annebyrne:
Examples of in-game play below:

screen shot 2017-02-07 at 16 23 43

screen shot 2017-02-07 at 16 25 04


LATaylor-guardian:
Just a general note, but if you plan on using these URLs in more than one place(which it looks like you are), then you should pull them out into a common object and reference that object in all of the places you wish to use a particular url. This means you only have one place to update if you need to change a url and is considered good practice.

shtukas:
:+1: Awesome, thanks @annebyrne

annebyrne:
@LATaylor-guardian I've refactored the default URL (share tile image) into a separate object, but haven't done so for the other URLs as they only appear elsewhere in tests, so hopefully should be sufficient for this once off.

mchv:
💯

PR #114, opened 2017-02-07
Initial message with custom briefing

tomrf1:
I left out the "Good morning" message that comes with the carousel, looks a bit weird without it


shtukas:
:+1:

PR #115, opened 2017-02-08
Deduplicate stories in custom briefing

tomrf1:
In case a story appears in both editors-picks and a topic's headlines. Keeping with the ordering:
1. editors-picks
2. topic1
3. editors-picks
4. topic2 || topic1
5. editors-picks

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/briefing/CustomBriefing.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/briefing/CustomBriefing.scala
@@ -24,14 +24,14 @@ object CustomBriefing {
         topic2Headlines <- user.briefingTopic2.map(topic2 => capi.getHeadlines(user.front, Topic.getTopic(topic2))).getOrElse(Future.successful(Nil))
       } yield {
 
-        //Interleave main headlines with topic stories
-        val stories: List[Content] = List(
-          headlines.headOption,
-          topic1Headlines.headOption,
-          headlines.lift(1),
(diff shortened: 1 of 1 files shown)


src/main/scala/com/gu/facebook_news_bot/briefing/CustomBriefing.scala
```diff
@@ -24,14 +24,14 @@ object CustomBriefing {
...
-          headlines.headOption,
-          topic1Headlines.headOption,
-          headlines.lift(1),
-          topic2Headlines.headOption orElse topic1Headlines.lift(1),
-          headlines.lift(2)
-        ).flatten
+        //Interleave main headlines with topic stories, and deduplicate
+        val ops =
+          appendNextNonDup(topic1Headlines) andThen
+          appendNextNonDup(headlines.tail) andThen
```
LATaylor-guardian: FYI A tail on an empty list will throw an `java.lang.UnsupportedOperationException`.

src/main/scala/com/gu/facebook_news_bot/briefing/CustomBriefing.scala
```diff
@@ -24,14 +24,14 @@ object CustomBriefing {
...
+          appendNextNonDup(if (topic2Headlines.nonEmpty) topic2Headlines else topic1Headlines.tail) andThen
```
LATaylor-guardian: And here.

PR #116, opened 2017-02-08
Move UpdateType to UserNoms table to facilitate querying when notifying users

annebyrne:


tomrf1:
This is what the end of the conversation currently looks like. The phrase "Great, thanks for playing!" sounds odd coming immediately after the sharing message.
We also discussed asking if they want to subscribe to morning briefings at this point instead, though we risk distracting them from the sharing.
screenshot_20170209-135446

tomrf1:
I've just pushed a commit to remove that final message, and instead add a 'Get Morning Briefing' quick reply after the share card. (This is the max number of chars I can fit in a quick reply)
screenshot_20170210-150500

tomrf1:
👍 please squash

PR #118, opened 2017-02-10
Use NLP to build capi queries

tomrf1:
### Existing functionality
We currently have a hardcoded set of topics, e.g. football, politics. These are mapped to specific sections/tags to ensure good results. If a user enters text containing a topic, the app returns a carousel with the links. If the user's input does not contain any of these hardcoded terms, it simply tells the user it doesn't understand.

### With this change
- If the user's input does not contain a topic then the app uses NLP to build a syntax tree from it (using http://stanfordnlp.github.io/CoreNLP/)
- Traverses the syntax tree looking for nouns, to build up a list of search terms.
- Uses CAPI's `q=` parameter to search for these terms. E.g. "Tell me about Obama and also Donald Trump" --> `/search?q=Obama+AND+%22Donald+Trump%22`
- The response to the user lists the terms it searched for, e.g. "I think you're asking me about obama and donald trump. Here are some articles that could be of interest." After the carousel, it asks "Was this helpful?" - and logs the user's response. It's good to show some humility with something with such variable results!

### Things to note:
- We already handle messages over 200 chars separately (by linking to readers' editor contact details), so it will never build large syntax trees. 
- Uses `order-by=relevance`, but then sorts the results by date to display the most recent. Otherwise the results can be very poor. It might be possible to tweak the [query decay params](https://github.com/guardian/content-api/blob/2cbab1c5ab45b36036866928f80cff48d686334e/concierge/src/main/scala/com.gu.contentapi.concierge/model/QueryGenerator.scala#L35) to get ES to do this for us, but we should not change existing behaviour in concierge.
- I've had to increase the ec2 from t2.micro to t2.small as the NLP library uses quite a lot of memory.

picture 5

picture 6

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/utils/Parser.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/utils/Parser.scala
@@ -1,0 +1,56 @@
+package com.gu.facebook_news_bot.utils
+
+import org.clulab.processors.corenlp.CoreNLPProcessor
+import org.clulab.struct.Tree
+
+import scala.collection.immutable.Queue
+
+object Parser {
+  private val nlpProcessor = new CoreNLPProcessor()
+
+  def warmUp = nlpProcessor.annotate("warm") //Avoids delaying the response to the first request
+
+  def getNouns(text: String): List[String] = {
+    val doc = nlpProcessor.annotate(text)
+
+    doc.sentences.toList.flatMap { sentence =>
+      sentence.syntacticTree.map { tree =>
+        getTerms(tree)
+      }
+    }.flatten
+  }
+
+  /**
+    * @param consecutiveNouns  nouns which should be considered a single term, e.g. "Donald Duck"
+    * @param overall           the final set of terms in this tree
+    */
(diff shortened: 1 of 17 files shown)


src/main/scala/com/gu/facebook_news_bot/utils/Parser.scala
```diff
@@ -0,0 +1,57 @@
...
+  private val nlpProcessor = new CoreNLPProcessor()
+
+  def warmUp = nlpProcessor.annotate("warm") //Avoids delaying the response to the first request
+
+  def getNouns(text: String): List[String] = {
+    val doc = nlpProcessor.annotate(text)
+
+    doc.sentences.toList.flatMap { sentence =>
+      sentence.syntacticTree.map { tree =>
+        //println(tree.toStringDepth(true))
```
LATaylor-guardian: Can be removed.

LATaylor-guardian:
Looks okay to me 👍  Will be interesting to see how the feedback from this.

mchv:
:diamond:

PR #121, opened 2017-02-13
Advertise Oscars Predictions Game to existing user base

annebyrne:
On 15th February, an invite to the Oscars game will be triggered to existing users with their morning briefing. An example of interaction is below:

picture 1

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
@@ -6,13 +6,14 @@ import com.gu.facebook_news_bot.briefing.MorningBriefingPoller._
 import com.gu.facebook_news_bot.models.{MessageToFacebook, User}
 import com.gu.facebook_news_bot.services.Facebook.FacebookMessageResult
 import com.gu.facebook_news_bot.services.{Capi, Facebook, FacebookEvents, SQSMessageBody}
-import com.gu.facebook_news_bot.state.MainState
+import com.gu.facebook_news_bot.state.{MainState, OscarsNomsStates}
 import com.gu.facebook_news_bot.state.StateHandler._
 import com.gu.facebook_news_bot.stores.UserStore
 import com.gu.facebook_news_bot.utils.Loggers._
 import com.gu.facebook_news_bot.utils.{JsonHelpers, Notifier, ResponseText, SQSPoller}
-import org.joda.time.{DateTime, DateTimeZone}
 import io.circe.generic.auto._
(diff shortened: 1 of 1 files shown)


src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
```diff
@@ -72,8 +69,15 @@ class MorningBriefingPoller(val userStore: UserStore, val capi: Capi, val facebo
       val variant = s"editors-picks-${user.front}"
       logBriefing(user.ID, variant)
 
-      MainState.getHeadlines(user, capi, Some(variant)) map { case (updatedUser, messages) =>
-        (updatedUser, morningMessage(updatedUser) :: messages)
+      if (user.oscarsNoms.isEmpty && SQSPoller.isDate(user.offsetHours, advertiseOscarNomsDate)) {
```
tomrf1: Just realised this needs to happen outside of this `getOrElse` block, as we want to ask people with the custom briefing as well

tomrf1:
👍

src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
```diff
@@ -76,5 +73,16 @@ class MorningBriefingPoller(val userStore: UserStore, val capi: Capi, val facebo
         (updatedUser, morningMessage(updatedUser) :: messages)
       }
     }
+
+    futureResult.flatMap {
+      case (briefingUser, message) =>
+        if (user.oscarsNoms.isEmpty && SQSPoller.isDate(user.offsetHours, advertiseOscarNomsDate)) {
+          for {
+            (updatedUser, oscarsInvite) <- OscarsNomsStates.InitialQuestionState.question(user, Some("Excited for the Oscars? Choose your favourite contenders and we'll let you know how your picks do on the night. Ready to vote?"))
```
LATaylor-guardian: What is the expected behaviour if the for comprehension short circuits here?

tomrf1: `question` here returns a future (though it actually just calls `Future.successful`). This should probably be a map rather than a for/yield

tomrf1: If you're wondering what happens when a future fails here, it bubbles up to [here](https://github.com/guardian/facebook-news-bot/blob/d2ae8cb26dda263aaac4954029c225a7041b45b9/src/main/scala/com/gu/facebook_news_bot/utils/SQSPoller.scala#L69)

PR #127, opened 2017-02-21
Deliver rolling-updates through night for Oscars winners notifications

annebyrne:


tomrf1:
This looks great!
Are we going to link to the live article? Or send an image?

annebyrne:
@tomrf1 I think suggestion to link to liveblog is a good idea. For the rolling updates I guess the real value is in knowing who won and how your predictions compared, rather than content?

I imagine we will be including content in links we send to morning-briefing users

PR #128, opened 2017-02-22
Restrict briefing topics

tomrf1:
This is a bug(?) I introduced with the NLP change, which models search terms as a special kind of `Topic` called `SearchTopic`.
The ability to add topics to your morning briefing is based on the available `Topic`s, but it should exclude arbitrary search terms!

I discovered this when I noticed that someone had chosen `hunks` as one of their morning briefing topics.


tsop14:
If we have content for a requested search term shouldn't we return that content?

mchv:
I love the feature actually 💎

tsop14:
HUNKS for all!

tomrf1:
@tsop14 the problem with allowing any search term is that the results might be old, and so they'll see the same articles about e.g. "hunks" every day. Which may, or may not, be desirable.

PR #132, opened 2017-02-27
Only include stories published in last 24 hours in custom briefing

tomrf1:
Stories must have a webPublicationDate from the last 24 hours to be included in the custom briefing. Some sections keep the same article at the top for days


LATaylor-guardian:
👍

PR #133, opened 2017-03-01
Excluded words list

tomrf1:
Taken from s3.
Certain words will now not be used when querying for content from capi.

I also removed oscars stuff from cloudformation as it was preventing me from updating it in CODE.

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/utils/BadWords.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/utils/BadWords.scala
@@ -1,0 +1,25 @@
+package com.gu.facebook_news_bot.utils
+
+import com.amazonaws.services.s3.AmazonS3Client
+import com.gu.cm.Mode
+import com.gu.facebook_news_bot.BotConfig
+
+object BadWords {
+  private val words: Set[String] = {
+    if (BotConfig.stage == Mode.Dev) Set("sausage")
+    else {
+      val s3Client: AmazonS3Client = new AmazonS3Client()
+      val stream = s3Client.getObject(BotConfig.badWordsBucket, BotConfig.badWordsFile).getObjectContent
+
+      val result = scala.io.Source.fromInputStream(stream).getLines.filter(_.nonEmpty).map(_.toLowerCase).toSet
+
+      stream.close()
+
+      result
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


src/main/scala/com/gu/facebook_news_bot/utils/BadWords.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import com.amazonaws.services.s3.AmazonS3Client
+import com.gu.cm.Mode
+import com.gu.facebook_news_bot.BotConfig
+
+object BadWords {
+  private val words: Set[String] = {
+    if (BotConfig.stage == Mode.Dev) Set("sausage")
+    else {
+      val s3Client: AmazonS3Client = new AmazonS3Client()
```
LATaylor-guardian: Can we create the client outside of the function please.

LATaylor-guardian:
👍

src/main/scala/com/gu/facebook_news_bot/utils/BadWords.scala
```diff
@@ -0,0 +1,21 @@
...
+      val stream = s3Client.getObject(BotConfig.badWordsBucket, BotConfig.badWordsFile).getObjectContent
```
mchv: Quoting `AWS` documentation:
```
Be extremely careful when using this method; the returned Amazon S3 object contains a direct stream of data from the HTTP connection. The underlying HTTP connection cannot be reused until the user finishes reading the data and closes the stream. Also note that if not all data is read from the stream then the SDK will abort the underlying connection, this may have a negative impact on performance. Therefore:

Use the data from the input stream in Amazon S3 object as soon as possible
Read all data from the stream (use GetObjectRequest.setRange(long, long) to request only the bytes you need)
Close the input stream in Amazon S3 object as soon as possible
If these rules are not followed, the client can run out of resources by allocating too many open, but unused, HTTP connections.
```
So don't forget to close the connection on the object after reading from it
```close
val obj = s3Client.getObject(BotConfig.badWordsBucket, BotConfig.badWordsFile)
...
...
obj.close()
```

LATaylor-guardian: Good spot!

tomrf1:
@mchv fixed!

PR #134, opened 2017-03-06
Add News Digest tile to Morning Briefing

annebyrne:
As an initial cut, we are delivering this to only UK readers. Clicking through on tile will take reader to Morning Briefing articles: https://www.theguardian.com/world/series/guardian-morning-briefing

If there is no fresh article, or the reader is outside UK, standard carousel is sent.

Once content in series is structured, plan is to give users option to consume content in bot.

Example seen on CODE: 
screen shot 2017-03-06 at 18 13 17

Note: Height of tiles is different due to absence of button, but have decided not to introduce one until we've added the content in bot feature

### DIFF
--- a/src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
+++ b/src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
@@ -3,14 +3,14 @@ package com.gu.facebook_news_bot.briefing
 import akka.actor.Props
 import com.gu.facebook_news_bot.BotConfig
 import com.gu.facebook_news_bot.briefing.MorningBriefingPoller._
-import com.gu.facebook_news_bot.models.{MessageToFacebook, User}
+import com.gu.facebook_news_bot.models.{Id, MessageToFacebook, User}
 import com.gu.facebook_news_bot.services.Facebook.FacebookMessageResult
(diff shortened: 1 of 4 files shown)


src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
```diff
@@ -50,17 +64,85 @@ class MorningBriefingPoller(val userStore: UserStore, val capi: Capi, val facebo
...
+  private def addMorningDigest(futureMaybeCarousel: Future[Option[MessageToFacebook.Message]], user: User): Future[Option[MessageToFacebook.Message]] = {
+    if (user.front == "uk") {
+      val futureMorningDigest: Future[Seq[Content]] = capi.getArticlesByTag("world/series/guardian-morning-briefing")
+
+      val today = DateTime.now(DateTimeZone.UTC)
+
+      val futureMaybeDigest: Future[Option[Content]] = futureMorningDigest.map(digest => checkDateContent(today, digest))
+
+      val carousel = for {
+        digestContent <- OptionT(futureMaybeDigest)
```
tomrf1: I've just realised this isn't quite right, as if there isn't a fresh article then we still want to return the original carousel, which won't happen currently

tomrf1:
👍

PR #136, opened 2017-03-17
Catch subscription-related phrases

tomrf1:
Some very basic handling of things I've noticed people saying frequently. It currently does a capi query when they're clearly trying to manage their subscription


shtukas:
:+1:

mchv:
💯

PR #137, opened 2017-03-27
Send now if user's timezone offset increased

tomrf1:
The bot gets the user's local time from facebook. It already updates the user's notification time when their UTC offset changes (though sometimes facebook takes a while to notice this).

Currently, when this happens it will not send a morning briefing until the next scheduled time. If the user's local clock has gone forward, it should send a notification now, as the user has now effectively missed today's notification.


src/main/scala/com/gu/facebook_news_bot/utils/Notifier.scala
```diff
@@ -37,10 +37,12 @@ object Notifier {
             if (fbUser.timezone == user.offsetHours) {
               Some(user)
             } else {
-              //User's timezone has changed - fix this now, but don't send briefing
+              //User's timezone has changed - fix this
               val updatedUser = changeNotificationTime(user, fbUser.timezone)
               updateUser(updatedUser, userStore)
-              None
+              //Only send messages if user's clock has gone forward
```
tsop14: Will you take this out after today?

tomrf1: No. This is the logic for checking if the user's UTC offset has changed. This is useful any time someone has a daylight savings change or travels across timezones (though facebook can be pretty slow to catch up).
The return value here indicates whether or not to go ahead and send the notification

tsop14:
👍

PR #138, opened 2017-06-05
Process batches of morning briefings in parallel

tomrf1:
The bot is now taking a long time to process pending morning briefings - it's been setting off a cloudwatch alarm recently. This is a small change to speed things up.

SQS allows you to take 10 items at a time. Previously, the app would wait until all 10 morning briefings have been processed and FB has responded before taking the next batch of 10 from SQS.
With this change, it can process more than one batch concurrently. The connection pool should support many concurrent connections to FB - it's unlikely it will need to buffer pending messages in memory.

Also, I've reinstated the throttler (this time as an akka-streams component rather than an actor), to avoid exceeding FB's rate limit.


LATaylor-guardian:
👍

PR #139, opened 2017-06-06
Improve questioning about topics

tomrf1:
1. When asking if user wants a custom topic in their briefing, offer "No" as the first quick reply, so they know they don't have to have them.
2. Don't ask about custom topics if they've already got two
3. Don't suggest a particular topic if they already have it


LATaylor-guardian:
👍

PR #141, opened 2017-06-26
Improve recording of uncontactables

tomrf1:
I noticed FB failing to return user data during morning briefings. We should always mark them as uncontactable, for more accurate subscriber counts.


LATaylor-guardian:
👍

PR #142, opened 2017-06-28
Unsubscribe users who have been uncontactable for 50 days

tomrf1:
This greatly reduces the number of requests we have to make to sqs/dynamo/FB


LATaylor-guardian:
I had the same query as @regiskuckaertz. If this is a 'yes' then 👍  I think it's useful to still have that information - even if in that state.

tomrf1:
@regiskuckaertz It unsubscribes them from the morning briefing. If a user's `notificationTime` fields are set to `-` then they are not subscribed (I wish I had used options instead, but I originally write this as a node app and used `-` as a special value).
There is a lambda that queues up morning briefings based on the `notificationTimeUTC`, so it will ignore users with a value of `-`.

PR #143, opened 2017-07-04
Add morningBriefingMaxConcurrentBatches config item

tomrf1:
This is the max number of batches to read from the queue before waiting for the FB requests to complete.
I want to try increasing it from 2, as it's taking a long time to get through the 8am briefings.


mchv:
👍  sounds sensible

PR #148, opened 2018-04-17
Final message

tomrf1:
We're decommissioning the Messenger bot. On a configured date, it will send out a final message to all users who are subscribed to the morning briefing at their normal briefing time. After this date, no more briefings will be sent out and we can switch it off.

uk
This is the message for UK users. For AU, the title of the first tile is 'Australia's Morning Mail', and for US/international only the last 2 tiles are sent.


src/main/scala/com/gu/facebook_news_bot/utils/FinalMessage.scala
```diff
@@ -0,0 +1,75 @@
...
+      title = Some("Android app"),
+      url = Some("https://play.google.com/store/apps/details?id=com.guardian"),
+      `type` = "web_url"
+    )
+  )
+
+  private val GuardianTodayEmailElement = MessageToFacebook.Element(
+    title = "The Guardian Today email",
+    subtitle = Some("Click here to get the headlines delivered to your inbox every morning"),
+    item_url = Some(s"https://www.theguardian.com/email-newsletters?$CampaignCode")
```
regiskuckaertz: no need for a parameter name like `utm_campaign` or something?

tomrf1: thanks!

regiskuckaertz:
giphy

src/main/scala/com/gu/facebook_news_bot/briefing/MorningBriefingPoller.scala
```diff
@@ -60,7 +61,26 @@ class MorningBriefingPoller(val userStore: UserStore, val capi: Capi, val facebo
...
+    val today = SQSPoller.getLocalDateTime(user.offsetHours).toLocalDate
+
+    if (today.isEqual(decommissionDate)) getFinalMessages(user)
+    else if (today.isBefore(decommissionDate)) getBriefingMessages(user)
+    else Future.successful(user, Nil)
+  }
+
+  private def getFinalMessages(user: User): Future[Result] = {
+    val messages = List(
+      MessageToFacebook.textMessage(user.ID, "Today we have taken a decision to close down the Guardian Facebook Messenger Bot.\n\nThe good news is that there are some great alternatives to get the best of the Guardian delivered to you every morning:"),
```
hosamaly: [Grammarly](grammarly.com) suggests _"decided"_ instead of _"have taken a decision"_.
> The phrase _taken a decision_ down may be unnecessarily wordy. Consider replacing the noun with a corresponding verb.

hosamaly: If I understand correctly, this message will be shown on the final day, so it shouldn't say "Today".

tomrf1: Indeed, I don't like the wording either. I think I'll write a new message :)

